PR #38, opened 2020-10-12
ci: zip binaries without bin/

afiune:
JIRA: ALLY-217


scottford-lw (approved):
👍  🙌

PR #39, opened 2020-10-13
feat: trigger initial compliance report automatically

afiune:
We are implementing a temporal workaround to mimic the behavior that is
codified inside the Lacework UI where, it automatically triggers an initial
report after the creation of Cloud Account Integrations, specifically for
Compliance (`AWS_CFG`, `GCP_CFG`, and `AZURE_CFG`)

We will remove this workaround when RAIN-13422 is done.

JIRA: ALLY-196


scottford-lw (approved):
👍  🙌  ⭐

PR #48, opened 2020-12-14
build: upgrade Go version to 1.15

dmurray-lacework:
Upgrade Go version to 1.15

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ orbs:
 executors:
   go-executor:
     docker:
-      - image: circleci/golang:1.14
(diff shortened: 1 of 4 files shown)


afiune (changes requested):
We need to update our pipeline to run on the same version, please update this line:

https://github.com/lacework/terraform-provider-lacework/blob/master/.circleci/config.yml#L18

afiune (changes requested):
Good catch about the macOS supported platform. We also need to remove this line:

https://github.com/lacework/terraform-provider-lacework/blob/master/scripts/release.sh#L21

afiune (approved):
tenor-199587965

PR #81, opened 2021-03-04
feat: retry mechanism for all cloud resources

afiune:
## User Story

As a Lacework & Terraform user,
I need to have stable and predictable executions,
So that I don't break my production deployments.

## Description

Users that have integrated our Terraform templates and providers into their CI/CD process are experiencing a number of API issues that are breaking their deployment processes/pipelines. We think that, when we introduce APIv2 into our Terraform provider, we will fix these issues but we cannot wait until this work is done in a couple of sprints, we need to have a solution sooner rather than later.

## Proposal Solution

We will implement a retry mechanism that will default to 5 tries with incremental intervals in between. This parameter will be configurable so that if a user wants to try 10 times or more, they can configure it.

## Acceptance Criteria

All cloud account resources should have implement this new retry mechanism, the resources we will implement it to are:

- [x] lacework_integration_aws_cfg
- [x] lacework_integration_aws_ct
- [x] lacework_integration_azure_al
- [x] lacework_integration_azure_cfg
- [x] lacework_integration_gcp_at
- [x] lacework_integration_gcp_cfg


scottford-lw (approved):
👍

website/docs/r/integration_aws_cfg.html.markdown
```diff
@@ -29,6 +29,7 @@ The following arguments are supported:
 * `name` - (Required) The AWS Config integration name.
 * `credentials` - (Required) The credentials needed by the integration. See [Credentials](#credentials) below for details.
 * `enabled` - (Optional) The state of the external integration. Defaults to `true`.
+* `retries` - (Optional) The number of attempts to create the external integration. Defaults to `5`.
```
afiune, line 32: @scottford-lw @dmurray-lacework is this a good text for the new argument?

dmurray-lacework (approved):

lacework/resource_lacework_integration_aws_cfg.go
```diff
@@ -34,6 +35,11 @@ func resourceLaceworkIntegrationAwsCfg() *schema.Resource {
 				Optional: true,
 				Default:  true,
 			},
+			"retries": {
```
afiune, line 38: @scottford-lw @dmurray-lacework Should this be `max_retries` instead? 🤔

dmurray-lacework: Yes, i think that better describes this parameters function.

dmurray-lacework: `max_retries` seems to be more the common nomenclature. At least from a quick search.
https://registry.terraform.io/providers/hashicorp/aws/latest/docs#max_retries

Not a sticking point for me and happy for this to go with either. But I do prefer max_retries

Issue #85, opened 2021-03-09
Changes to lacework_integration_ecr Name field have no effect


alexeiser:
If a Lacework ECR resource has been created, or imported - and the name value changes no effect is shown on the plan (as if ignore_changes was set).

Example resource config, with a changed name.
```
resource "lacework_integration_ecr" "lacework-ecr-integration" {
  name              = "new name"
  registry_domain   = "${data.aws_caller_identity.current.account_id}.dkr.ecr.${var.region}.amazonaws.com"
  access_key_id     = aws_iam_access_key.lacework-ecr-access_key.id
  secret_access_key = aws_iam_access_key.lacework-ecr-access_key.secret
}
```

Existing object in Lacework:
```
 {
      "INTG_GUID": "GRAILBIO_XXXXX",
      "NAME": "lacework-ecr-integration",
      "CREATED_OR_UPDATED_TIME": "2020-Dec-11 19:48:39 UTC",
      "CREATED_OR_UPDATED_BY": "XXXX",
      "TYPE": "CONT_VULN_CFG",
      "ENABLED": 1,
      "STATE": {
        "ok": true,
        "lastUpdatedTime": "2021-Mar-09 03:16:50 UTC",
        "lastSuccessfulTime": "2021-Mar-09 03:16:50 UTC"
      },
      "IS_ORG": 0,
      "DATA": {
        "ACCESS_KEY_CREDENTIALS": {
          "ACCESS_KEY_ID": "XXXXXXXXX"
        },
        "AWS_AUTH_TYPE": "AWS_ACCESS_KEY",
        "REGISTRY_TYPE": "AWS_ECR",
        "REGISTRY_DOMAIN": "XXXXXXXX.dkr.ecr.us-west-2.amazonaws.com",
        "LIMIT_BY_TAG": "*",
        "LIMIT_BY_LABEL": "*",
        "LIMIT_BY_REP": "",
        "LIMIT_NUM_IMG": 5
      },
      "TYPE_NAME": "CONT_VULN_CFG"
    },
```

Terraform plan result
```

No changes. Infrastructure is up-to-date.
```

afiune:
Hi @alexeiser 👋🏽!! Thank you for your feedback, we are going to try to replicate this bug in our environment, could you please help us by providing the version of Terraform you are running on?

afiune:
Tested on Terraform 14 and seems to be working fine:
[...]

PR #86, opened 2021-03-13
fix: lacework_integration_ecr from SCHEMA changes and IAM Role support

afiune:
This PR is a bug-fix since our customers are experiencing failures:

Closes https://github.com/lacework/terraform-provider-lacework/issues/85
 
But it is also a feature! 🙌🏽🎉⭐

## feat: lacework_integration_ecr support for IAM Roles

We now support integrating ECR using IAM Roles instead of keys, though,
we will continue to support access keys.

To create an ECR integration with the new IAM role authentication method:
```hcl
provider "lacework" {}

resource "lacework_integration_ecr" "iam_role" {
  name            = "ECR using IAM Role"
  registry_domain = "YourAWSAccount.dkr.ecr.YourRegion.amazonaws.com"
  credentials {
    role_arn    = "arn:aws:iam::1234567890:role/lacework_iam_example_role"
    external_id = "12345"
  }
}
```

To create an ECR integration with the old authentication method, using
Access Keys:
```hcl
provider "lacework" {}

resource "lacework_integration_ecr" "access_key" {
  name            = "ECR using Access Keys"
  registry_domain = "YourAWSAccount.dkr.ecr.YourRegion.amazonaws.com"
  credentials {
    access_key_id     = "AWS123abcAccessKeyID"
    secret_access_key = "AWS123abc123abcSecretAccessKey0000000000"
  }
}
```


dmurray-lacework (approved):

lacework/resource_lacework_integration_ecr.go
```diff
@@ -1,13 +1,51 @@
...
+	}
+
+	for _, integration := range response.Data {
+		if integration.IntgGuid == d.Id() {
+			log.Printf("[INFO] Integration found with guid: %v\n", integration.IntgGuid)
+
+			// @afiune this field is important since it tells the resource which API functions to use
+			// we will extract it from the raw integration response so that the READ functions populate
+			// the rest of the fiels
+			if awsAuthType, ok := integration.Data["AWS_AUTH_TYPE"]; ok {
```
afiune, line 29: We don't use the regular integration importer func since for ECR integrations we must know what the Authentication method is to use the appropriate GO-SDK functions.

lacework/resource_lacework_integration_ecr.go
```diff
@@ -33,14 +71,32 @@ func resourceLaceworkIntegrationEcr() *schema.Resource {
...
 				Required: true,
 			},
-			"access_key_id": {
-				Type:     schema.TypeString,
+			"credentials": {
+				Type:     schema.TypeList,
+				MaxItems: 1,
 				Required: true,
-			},
-			"secret_access_key": {
```
afiune, line 40: We will stop supporting these arguments at the root of the resource and therefore we will bump the version to `0.2.0`. Thought should we be bumping to `1.0.0`since this change is not backward compatible?

PR #102, opened 2021-05-12
build: Update go version to 1.16

dmurray-lacework:
Upgrade Go version


afiune (approved):
tenor-229661247

PR #113, opened 2021-06-15
fix: deprecate docker_v2 attributes tag/label limits

afiune:
We will remove these fields in v1.0, for now, we are just marking them
as deprecated so that our users can remove them.

JIRA: ALLY-533


scottford-lw (approved):
👍 thanks @afiune

dmurray-lacework (approved):

afiune:
We decided NOT to deprecate these attributes, but instead, update them like the rest of the Container Registries Integrations.

PR #114, opened 2021-06-15
feat: add support for multiple tags and labels

afiune:
We are introducing a new feature that will allow our users to limit
scans not only by a single tag or a single label, but now they will be
able to pass multiple tags and labels.

We are deprecating old attributes that were of type `String` in favor of
new attributes of type `List(string)`.

Moved Attributes:
* `limit_by_tag` -> `limit_by_tags`
* `limit_by_label` -> `limit_by_labels`
* `limit_by_repos` -> `limit_by_repositories`

If a user is using a deprecated attribute, we will prompt the following message:

Screen Shot 2021-06-14 at 5 15 41 PM

If a user tries to use both attributes, the old and new one, we will prompt the following message:

Screen Shot 2021-06-14 at 4 45 26 PM

JIRA: ALLY-533


lacework/utils.go
```diff
@@ -17,11 +28,32 @@ func castAndTransformStringArray(iArray []interface{}, f func(string) string) []
 	return a
 }
 
-// turn a string array into an instance array
-func castStringArrayToInterface(strs []string) []interface{} {
+// turn a string slice into an instance slice
+func castStringSliceToInterface(strs []string) []interface{} {
```
dmurray-lacework, line 32: I don't see any usages of this func

afiune: oh right, back in the days, I thought this could be a good way to also transform/validate fields,
but to be honest, we haven’t used it :sweat_smile: I am happy to remove it if it is too complex.

lacework/resource_lacework_integration_ecr.go
```diff
@@ -347,16 +422,32 @@ func resourceLaceworkIntegrationEcrUpdate(d *schema.ResourceData, meta interface
 }
 
 func resourceLaceworkIntegrationEcrUpdateWithIAMRole(d *schema.ResourceData, lacework *api.Client) error {
+
```
dmurray-lacework, line 425: nit: remove line

lacework/resource_lacework_integration_ecr.go
```diff
@@ -401,16 +492,32 @@ func resourceLaceworkIntegrationEcrUpdateWithIAMRole(d *schema.ResourceData, lac
 }
 
 func resourceLaceworkIntegrationEcrUpdateWithAccessKey(d *schema.ResourceData, lacework *api.Client) error {
+
```
dmurray-lacework, line 495: nit: remove line

lacework/resource_lacework_integration_ecr.go
```diff
@@ -558,16 +665,32 @@ func resourceLaceworkIntegrationEcrCreateWithAccessKey(d *schema.ResourceData, l
 }
 
 func resourceLaceworkIntegrationEcrCreateWithIAMRole(d *schema.ResourceData, lacework *api.Client) error {
+
```
dmurray-lacework, line 668: nit: line

lacework/utils.go
```diff
@@ -17,11 +28,32 @@ func castAndTransformStringArray(iArray []interface{}, f func(string) string) []
...
 	arr := make([]interface{}, len(strs))
 	for i, str := range strs {
 		arr[i] = str
 	}
 	return arr
 }
+
+// extract an attribute from the provided ResourceData and convert it into a map of strings
+// with string keys
+func castAttributeToStringKeyMapOfStrings(d *schema.ResourceData, attr string) map[string]string {
```
dmurray-lacework, line 42: Could we add a test for this function aswell?

afiune:
tenor-251613883

PR #121, opened 2021-07-20
test: e2e testing with terratest

dmurray-lacework:
First e2e test of the terraform-provider-lacework resources using terratest.
- New e2e test for datadog alert channel resource
- `make mirror-lacework-provider` command to use local binary on ci when running e2e terraform tests 

JIRA: https://lacework.atlassian.net/browse/ALLY-548


examples/resource_lacework_alert_channel_datadog/main.tf
```diff
@@ -1,7 +1,21 @@
+terraform {
+  required_providers {
+    lacework = {
+      source = "lacework/lacework"
+      version = "0.5.0"
```
afiune, line 5: How can we source the binary built locally? 🤔  -- This is testing what is in the Terraform Registry, right?

integration/integration.go
```diff
@@ -0,0 +1,36 @@
+package integration
+
+import (
+	"github.com/lacework/go-sdk/api"
+	"log"
+	"os"
+	"strings"
+)
+
+var LwIntegrationClient *api.Client
```
afiune, line 10: nit: maybe better name?
```suggestion
var LwClient *api.Client
```

afiune (approved):
tenor-172922452

PR #124, opened 2021-07-26
feat: new lacework_alert_channel_email resource

afiune:
# User Story

As a Lacework customer and a Terraform user,
I want to be able to use Terraform to configure Email Alert Channel integrations in my Lacework account,
So I can configure and maintain multiple alerts as code.

Integration Type: `EMAIL_USER`

## Example
```hcl
provider "lacework" {}

resource "lacework_alert_channel_email" "example" {
  name       = "notify team ABC"
  recipients = ["foo@example.com"]
}
```

JIRA: https://lacework.atlassian.net/browse/ALLY-102


GNUmakefile
```diff
@@ -14,7 +14,6 @@ GOFLAGS=-mod=vendor
 CGO_ENABLED?=0
 PACKAGENAME?=terraform-provider-lacework
 VERSION=$(shell cat VERSION)
-LOCAL_PROVIDERS="$$HOME/.terraform.d/plugins"
```
afiune, line 17: This is exactly the same path as `$DIR`

integration/integration.go
```diff
@@ -15,7 +16,10 @@ func init() {
 
 func lwTestCLient() (lw *api.Client) {
 	lw, err := api.NewClient(os.Getenv("LW_ACCOUNT"),
-		api.WithApiKeys(os.Getenv("LW_API_KEY"), os.Getenv("LW_API_SECRET")))
+		api.WithApiKeys(os.Getenv("LW_API_KEY"), os.Getenv("LW_API_SECRET")),
+		api.WithSubaccount(os.Getenv("LW_SUBACCOUNT")),
```
afiune, line 20: This works whether or not there is a subacount. 👍🏽

integration/integration.go
```diff
@@ -28,8 +32,8 @@ func GetIntegrationName(result string) string {
 	id := strings.Split(resultSplit[1], "]")[0]
 
 	res, err := LwClient.Integrations.Get(id)
-	if err != nil {
-		log.Fatalf("Unable to find integration id: %v", id)
+	if err != nil || len(res.Data) == 0 {
```
afiune, line 35: I found a panic when we don't fine any integration and we try to access `.Data[0].Name`

scripts/mirror-provider.sh
```diff
@@ -1,16 +1,21 @@
...
-            filesystem_mirror {
-              path    = "/root/.terraform.d/plugins"
-              include = ["lacework/lacework"]
-            }
-            direct {
-              exclude = ["lacework/lacework"]
-            }
-          }
+    filesystem_mirror {
+        path    = "$HOME/.terraform.d/plugins"
```
afiune, line 12: **It took me a while to figure this one out** 

tenor-133131062

dmurray-lacework (approved):

PR #129, opened 2021-07-28
ci: clean left over files during integration test

afiune:


afiune:
tenor-20572991

PR #133, opened 2021-08-02
feat: Run Test Alert Channel during TF Create

dmurray-lacework:
https://lacework.atlassian.net/browse/ALLY-331

*Example of adding Test Integration to TF Create of Alert Channels

Alert Channel resources run "Test Integration" `v2/AlertChannels/IntGuid/test` during Terraform Create.
The resource is deleted upon failing the Test.

Skip running the Test Integration with `test_integration` flag

```
resource "lacework_alert_channel_datadog" "example" {
  name            = var.channel_name
  ...
  test_integration    = true
}
```

Signed-off-by: Darren Murray <darren.murray@lacework.net>

PR #138, opened 2021-08-06
refactor: test Alert Channels on create and modify

afiune:
PR https://github.com/lacework/terraform-provider-lacework/pull/133 only tested alert channels after creation, but we need to test them
also when users modify them.

Corrected logging that always printed Datadog Integration.


afiune:
tenor-86792930

Issue #147, opened 2021-08-17
Unable to manage Alert Channels at the Organization level


afiune:
As a Lacework customer that is enrolled into a Lacework organization,
I want to be able to manage my Alert Channels at the Organization level,
So I can manage Alert Channels shared across Lacework sub-accounts.

## Bug Description
Today, if any user try to manage an Alert Channel at the Organization level,
for example:
```hcl
provider "lacework" {
  organization = true
}

resource "lacework_alert_channel_slack" "ops" {
  name = "ops critical alerts"
  slack_url = "https://hooks.slack.com/services/ABCD/12345/abcd1234"
}
```
The Lacework provider doesn't create (manage) this alert channel at the organization level,
but rather, at the sub-account level.

## Proposal Solution
The problem we are facing is the fact that we are migrating the Lacework provider
from API v1 to API v2 and none of the Alert Channels have been migrated to v2.

APIv1 doesn't allow access to organization level data sets, so the solution is to migrate
all Alert channels to APIv2.

PR #148, opened 2021-08-17
fix: migrate Slack Alert Channels to use API v2

afiune:
As a Lacework customer that is enrolled in a Lacework organization,
I want to be able to manage my Alert Channels at the Organization level,
So I can manage Alert Channels shared across Lacework sub-accounts.

# Description
We are migrating Alert Channels to use APIv2 so that our customers can
manage them at the organization level, we are starting with the Slack alert
channel and moving down to move them all.

# Example
Customers should now be able to run the following Terraform code:
```hcl
terraform {
  required_providers {
    lacework = {
      source = "lacework/lacework"
    }
  }
}

provider "lacework" {
  organization = true
}

resource "lacework_alert_channel_slack" "example" {
  name      = "Slack Alert Channel Example"
  slack_url = "https://hooks.slack.com/services/ABCD/12345/abcd1234"
}
```

Contributes to https://github.com/lacework/terraform-provider-lacework/issues/147


afiune:
# Gonna Go Solo on This One
tenor-214015443

PR #149, opened 2021-08-18
chore: fix Go deps

afiune:
Just ran:
```
make go-vendor
```


afiune:
tenor-162676581

PR #167, opened 2021-09-23
fix: return empty array when casting limit_by fields

dmurray-lacework:
issue: https://lacework.atlassian.net/browse/ALLY-674

The function `castAndTransformStringSlice` was returning nil when an empty array is passed in. 


afiune (approved):
tenor-122658840

afiune:
Just made an update to the title to avoid writing `fix` twice.

PR #179, opened 2021-09-28
feat: Add new field non_os_package_support to GAR resource

dmurray-lacework:
Issue: https://lacework.atlassian.net/browse/ALLY-664

Description:
Add new field non_os_package_support for gar terraform resource

Signed-off-by: Darren Murray darren.murray@lacework.net


examples/resource_lacework_integration_gar/variables.tf
```diff
@@ -18,3 +18,7 @@ variable "private_key" {
   type      = string
   sensitive = true
 }
+variable "non_os_packages" {
```
vatasha, line 21: nit: let's name the variable `non_os_package_support` to be consistent.

integration/integration.go
```diff
@@ -57,6 +57,18 @@ func GetIntegrationName(result string) string {
 	return res.Data[0].Name
 }
 
+func GetContainerRegistryIntegration(result string) api.ContainerRegIntegration {
```
vatasha, line 60: q: why are we doing this? I'm concerned about us splitting a string and the string itself changing. This feels a little brittle.

integration/resource_lacework_integration_gar_test.go
```diff
@@ -18,9 +18,10 @@ func TestIntegrationGARCreate(t *testing.T) {
 		terraformOptions := terraform.WithDefaultRetryableErrors(t, &terraform.Options{
 			TerraformDir: "../examples/resource_lacework_integration_gar",
 			Vars: map[string]interface{}{
-				"client_id":      gcreds.ClientID,
```
vatasha, line 21: nit: I think underscores for field names is very pythonic and not idiomatic go. I would use camelcase and not snakecase.

lacework/resource_lacework_integration_gar.go
```diff
@@ -192,11 +198,12 @@ func resourceLaceworkIntegrationGarCreate(d *schema.ResourceData, meta interface
 	data := api.NewContainerRegistry(d.Get("name").(string),
```
vatasha, line 198: I noticed we're doing a lot of type assertions without changing for an error. I think we should be checking for errors for all of these assertions.

vatasha (changes requested):
Nothing major here but I'd love to have a conversation about error handling before this gets merged.

examples/resource_lacework_integration_gar/variables.tf
vatasha:
```suggestion
variable "non_os_package_support" {
```

examples/resource_lacework_integration_gar/main.tf
```diff
@@ -17,6 +17,7 @@ resource "lacework_integration_gar" "example" {
     private_key_id = var.private_key_id
     private_key    = var.private_key
   }
+  non_os_package_support = var.non_os_packages
```
vatasha, line 20:
```suggestion
  non_os_package_support = var.non_os_package_support
```

integration/resource_lacework_integration_gar_test.go
```diff
@@ -18,9 +18,10 @@ func TestIntegrationGARCreate(t *testing.T) {
...
-				"client_email":   gcreds.ClientEmail,
-				"private_key_id": gcreds.PrivateKeyID,
+				"client_id":       gcreds.ClientID,
+				"client_email":    gcreds.ClientEmail,
+				"private_key_id":  gcreds.PrivateKeyID,
+				"non_os_packages": false,
```
vatasha, line 24:
```suggestion
				"non_os_package_support": false,
```

lacework/resource_lacework_integration_gar.go
vatasha: Update: Discussed this and understand that the schema validates the type so there is no need for checking errors on the type assertions.

vatasha (approved):

integration/resource_lacework_integration_gar_test.go
dmurray-lacework: We're overriding the terraform variables here. https://www.terraform-best-practices.com/naming

integration/integration.go
dmurray-lacework: It certainly could be improved. The string being split is the terraform output, which is consistent. If there was no output or no id in the output the tf step would have failed. But I can look at pattern matching instead.

afiune (approved):
tenor-157862798

afiune:
@dmurray-lacework the pipeline is complaining about:
```
Executing command: make test                                                                                                                                          
==> Checking that code complies with gofmt requirements...                                                                                                            
gofmt needs running on the following files:                                                                                                                           
./integration/resource_lacework_integration_gar_test.go                                                                                                               
You can use the command: `make fmt` to reformat code.                                                                                                                 
make: *** [GNUmakefile:113: fmtcheck] Error 1                                                                                                                         
```

PR #180, opened 2021-09-30
test: fix integration tests

dmurray-lacework:

***Description:***
Fix integration tests


afiune (approved):
We should look into why this happened...

PR #189, opened 2021-10-05
fix: Migrate Datadog alert channel to API v2

vatasha:
---
name: Migrate Datadog alert channel to API v2
about: fix: Migrate Datadog alert channel to API v2
---

***Issue***: https://lacework.atlassian.net/browse/ALLY-639?atlOrigin=eyJpIjoiMWUzZmU0NGQ2M2NmNDVhYzg1NjZiMGNjZTA4MDVjM2YiLCJwIjoiaiJ9

***Description:***
Migrate Datadog alert channel to use V2 of the API. This is a part of a bigger effort to allow customers to manager alert channels at the org level rather than being limited to the sub-account level.


lacework/resource_lacework_alert_channel_datadog.go
```diff
@@ -111,80 +111,70 @@ func resourceLaceworkAlertChannelDatadogCreate(d *schema.ResourceData, meta inte
...
 		lacework = meta.(*api.Client)
-		datadog  = api.NewDatadogAlertChannel(d.Get("name").(string),
-			api.DatadogChannelData{
-				DatadogSite:    site,
-				DatadogService: service,
-				ApiKey:         d.Get("api_key").(string),
+		datadog  = api.NewAlertChannel(d.Get("name").(string),
+			api.DatadogAlertChannelType,
+			api.DatadogDataV2{
+				DatadogSite: string(site),
```
dmurray-lacework, line 117: We wouldn't need to cast to string again if we set the DatadogSite and Type as `datadogSite`  and `datadogService` in the go-sdk
https://github.com/lacework/go-sdk/pull/570/files#r723033234

An alternative would be to change datadog_site and datadog_type in the schema. But I think having the protection for only valid types would be preferable.

lacework/resource_lacework_alert_channel_datadog.go
```diff
@@ -194,11 +184,12 @@ func resourceLaceworkAlertChannelDatadogUpdate(d *schema.ResourceData, meta inte
...
+				DatadogType: string(service),
```
dmurray-lacework, line 191: Same as comment above about setting the types in go-sdk

dmurray-lacework (changes requested):

integration/resource_lacework_alert_channel_datadog_test.go
```diff
@@ -22,7 +29,8 @@ func TestDatadogAlertChannelCreate(t *testing.T) {
 
 	// Update Datadog Alert Channel
 	terraformOptions.Vars = map[string]interface{}{
-		"channel_name": "Datadog Alert Channel Updated"}
+		"channel_name": "Datadog Alert Channel Updated",
+		"api_key":      apiKey}
```
afiune, line 33: Could we do two things:
1. update the fields that we have above to verify that "updates" are working fine
2. assert more fields and not only the integration name, things like the channel name, site, service and API key
```suggestion
		"api_key":      "updated-fake-key"}
```

lacework/resource_lacework_alert_channel_datadog.go
```diff
@@ -111,80 +111,70 @@ func resourceLaceworkAlertChannelDatadogCreate(d *schema.ResourceData, meta inte
...
-
-	d.SetId("")
+	d.Set("name", response.Data.Name)
+	d.Set("intg_guid", response.Data.IntgGuid)
+	d.Set("enabled", response.Data.Enabled == 1)
+	d.Set("created_or_updated_time", response.Data.CreatedOrUpdatedTime)
+	d.Set("created_or_updated_by", response.Data.CreatedOrUpdatedBy)
+	d.Set("type_name", response.Data.Type)
+	d.Set("org_level", response.Data.IsOrg == 1)
+	d.Set("api_key", response.Data.Data.ApiKey)
```
afiune, line 172: The API server will never return this key, we should not add this Set statement. (which it wasn't here before)
```suggestion
```

afiune: Definitely a blocker 👆🏽

PR #192, opened 2021-10-07
fix: Migrate VictorOps alert channel(v2)

dmurray-lacework:
Requires [go-sdk change](https://github.com/lacework/go-sdk/pull/573) merged and deps updated

***Issue***: https://lacework.atlassian.net/browse/ALLY-650 https://github.com/lacework/terraform-provider-lacework/issues/147

***Description:***
Migrate webhook alert channel to use api v2

### DIFF
--- a/examples/resource_lacework_alert_channel_victorops/main.tf
+++ b/examples/resource_lacework_alert_channel_victorops/main.tf
@@ -19,9 +19,9 @@ resource "lacework_alert_channel_victorops" "example" {
 }
 
 output "channel_name" {
-  value = var.channel_name
+  value = lacework_alert_channel_victorops.example.name
 }
 
 output "webhook_url" {
-  value = var.webhook_url
+  value = lacework_alert_channel_victorops.example.webhook_url
 }


examples/resource_lacework_alert_channel_victorops/main.tf
```diff
@@ -1,6 +1,27 @@
...
+  webhook_url = var.webhook_url
+  // test_integration input is used in this example only for testing
+  // purposes, it help us avoid sending a "test" request to the
+  // system we are integrating to. In production, this should remain
+  // turned on ("true") which is the default setting
+  test_integration = false
 }
+
+output "channel_name" {
+  value = var.channel_name
```
afiune, line 22: There is no sense in having an output to point to a variable, we are not testing anything there. 

Point them to the resource:
```suggestion
  value = lacework_alert_channel_victorops.example.name
```

examples/resource_lacework_alert_channel_victorops/main.tf
```diff
@@ -1,6 +1,27 @@
...
+}
+
+output "webhook_url" {
+  value = var.webhook_url
```
afiune, line 26: There is no sense in having an output to point to a variable, we are not testing anything there. 

Point them to the resource:
```suggestion
  value = lacework_alert_channel_victorops.example.webhook_url
```

examples/resource_lacework_alert_channel_victorops/main.tf
afiune: cc @vatasha 👆🏽

examples/resource_lacework_alert_channel_victorops/main.tf
afiune: cc @vatasha 👆🏽

Issue #181, opened 2021-09-30
bug: if lacework_alert_channel_slack fails, provider has resource in state, but it does not exist in Lacework


timarenz:
When the initial creation of the `lacework_alert_channel_slack` fails, for example, because the `slack_url` is wrong and `test_integration` is set to `true` (which is the default) the Lacework provider thinks the alert channel is created in the Lacework platform but in reality it isn't.

Based on this subsequent terraform operations (apply and destroy) fail, as the Lacework API response with an `[404] Not found`.

Steps to reproduce:

main.tf
```hcl
provider "lacework" {

}

resource "lacework_alert_channel_slack" "ops_critical" {
  name      = "OPS Critical Alerts"
  slack_url = "https://hooks.slack.com/services/ABCD/12345/abcd1234"
}
```

Initial apply
```bash
$ terraform apply -auto-approve

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # lacework_alert_channel_slack.ops_critical will be created
  + resource "lacework_alert_channel_slack" "ops_critical" {
      + created_or_updated_by   = (known after apply)
      + created_or_updated_time = (known after apply)
      + enabled                 = true
      + id                      = (known after apply)
      + intg_guid               = (known after apply)
      + name                    = "OPS Critical Alerts"
      + org_level               = (known after apply)
      + slack_url               = "https://hooks.slack.com/services/ABCD/12345/abcd1234"
      + test_integration        = true
      + type_name               = (known after apply)
    }

Plan: 1 to add, 0 to change, 0 to destroy.
lacework_alert_channel_slack.ops_critical: Creating...
╷
│ Error:
│   [POST] https://lwintadriandaniels.fra.lacework.net/api/v2/AlertChannels/LWINTTIM_500FC239AFE4CAF64C4704382EC43111D01A57AC16E94B7/test
│   [404] no_team
│
│   with lacework_alert_channel_slack.ops_critical,
│   on main.tf line 27, in resource "lacework_alert_channel_slack" "ops_critical":
│   27: resource "lacework_alert_channel_slack" "ops_critical" {
│
╵
```

Statefile
```json
{
  "version": 4,
  "terraform_version": "1.0.7",
  "serial": 33,
  "lineage": "a75494b9-103b-0f78-5a4d-b9b49b219e10",
  "outputs": {},
  "resources": [
    {
      "mode": "managed",
[...]

PR #199, opened 2021-10-12
fix: remove alert channel properly if test fails

afiune:
***Issue***:
* https://lacework.atlassian.net/browse/ALLY-688
* https://github.com/lacework/terraform-provider-lacework/issues/181

***Description:***
When users create new alert channels, by default we test the
integrations, when it fails, we remove it from their Lacework account
but we forgot to remove it from the Terraform state.

We are fixing
https://github.com/lacework/terraform-provider-lacework/issues/181 by
removing the alert channel from the Terraform state.

***Additional Info:***
To test this, go into any example in our `examples/` folder, switch the flag `test_integration`
to `true` and run Terraform. On a second `terraform apply` you should not see the message:
```
│ Error:
│   [POST] https://account.lacework.net/api/v2/AlertChannels/ABC_500FC239AFE4CAF64C4704382EC43111D01A57AC16E94B7/test
│   [404] no_team
```


dmurray-lacework (approved):
1 typo to fix

integration/integration.go
```diff
@@ -165,3 +155,12 @@ func GetAlertChannelProps(result string) api.AlertChannelResponse {
 	}
 	return data
 }
+
+func GetIDFromTerraResults(result string) string {
```
vatasha, line 159: Nice!

lacework/alert_channel_importer.go
```diff
@@ -12,10 +12,11 @@ import (
...
 // if the test is not successful, it will remove the alert channel (rollback)
-func VerifyAlertChannelAndRollback(id string, lacework *api.Client) error {
-	if err := lacework.V2.AlertChannels.Test(id); err != nil {
+func VerifyAlertChannelAndRollback(d *schema.ResourceData, lacework *api.Client) error {
+	if err := lacework.V2.AlertChannels.Test(d.Id()); err != nil {
+		defer d.SetId("")
 		// rollback terraform create upon error testing integration
-		if deleteErr := lacework.V2.AlertChannels.Delete(id); deleteErr != nil {
+		if deleteErr := lacework.V2.AlertChannels.Delete(d.Id()); deleteErr != nil {
 			return errors.Wrapf(deleteErr, "Unable to rollback changes: %v", err)
```
vatasha, line 20: Let's use a domain specific error here as presented in the article above.

vatasha (changes requested):
I would like to see the way we do error checking done a little differently if you're open to my suggestions :)

lacework/resource_lacework_alert_channel_victorops.go
```diff
@@ -193,31 +193,4 @@ func resourceLaceworkAlertChannelVictorOpsDelete(d *schema.ResourceData, meta in
 
 	log.Printf("[INFO] Deleted %s integration with guid %s\n", api.VictorOpsAlertChannelType, d.Id())
 	return nil
-}
-
-func validateVictorOpsAlertChannelResponse(response *api.VictorOpsAlertChannelResponse) error {
```
afiune, line 198: I think we forgot to remove this code. cc @dmurray-lacework I removed it for you 😉

PR #202, opened 2021-10-12
ci: fix release pipeline

afiune:


afiune:
I agree with that, I have created https://lacework.atlassian.net/browse/ALLY-707 to deal with this issue.

PR #206, opened 2021-10-15
fix: Migrate ServiceNow alert channel(v2)

dmurray-lacework:
***Issue***: https://lacework.atlassian.net/browse/ALLY-648 https://github.com/lacework/terraform-provider-lacework/issues/147

***Description:***
Migrate Service Now alert channel to use api v2


integration/resource_lacework_integration_service_now_test.go
```diff
@@ -0,0 +1,50 @@
...
+			"custom_template_file": customTemplate,
+		},
+	})
+	defer terraform.Destroy(t, terraformOptions)
+
+	// Create new ServiceNowRest Alert Channel
+	create := terraform.InitAndApplyAndIdempotent(t, terraformOptions)
+	created := GetAlertChannelProps(create)
+
+	assert.Equal(t, "Service Now Alert Channel Example", created.Data.Name)
```
afiune, line 30: can we assert all the above values before the next apply? Something like we do below but here:

integration/resource_lacework_integration_service_now_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	actualName := terraform.Output(t, terraformOptions, "channel_name")
+	actualUrl := terraform.Output(t, terraformOptions, "instance_url")
+	actualUsername := terraform.Output(t, terraformOptions, "username")
+	actualTemplate := terraform.Output(t, terraformOptions, "custom_template_file")
+	assert.Equal(t, "Service Now Alert Channel Updated", updated.Data.Name)
+	assert.Equal(t, "Service Now Alert Channel Updated", actualName)
+	assert.Equal(t, "https://dev123.service-now.com", actualUrl)
+	assert.Equal(t, "snow-user", actualUsername)
+	assert.Equal(t, customTemplate, actualTemplate)
```
afiune, line 47: I think we should also be asserting what is inside `update.Data.Data["field"]`

integration/resource_lacework_integration_service_now_test.go
```diff
@@ -0,0 +1,73 @@
...
+		data["customTemplateFile"])
+	assert.Equal(t, "snow-user", data["userName"])
+
+	assert.Equal(t, "https://dev123.service-now.com", actualUrl)
+	assert.Equal(t, "snow-user", actualUsername)
+	assert.Equal(t, customTemplate, actualTemplate)
+
+	// Update ServiceNowRest Alert Channel
+	terraformOptions.Vars = map[string]interface{}{
+		"channel_name": "Service Now Alert Channel Updated"}
```
afiune, line 49: Can we inject new fields for all the variables/inputs, that way we can test that they all changed correctly.

afiune (changes requested):

PR #237, opened 2021-11-29
feat(resource): New lacework_report_rule

dmurray-lacework:
***Issue***: https://lacework.atlassian.net/browse/ALLY-756

***Description:***
Add new Terraform resource for report rules

Usage: 
``` hcl
resource "lacework_report_rule" "example" {
  name            = var.name
  description     = var.description
  enabled         = true
  severities      = var.severities
  resource_groups = var.resource_groups
  email_alert_channels = var.channels

  aws_compliance_reports {
    pci    = true
    cis_s3 = true
  }

  weekly_snapshot = var.snapshot
}
```


examples/resource_lacework_report_rule/main.tf
```diff
@@ -0,0 +1,101 @@
...
+  daily_compliance_reports {
+    aws_cloudtrail = var.daily_cloudtrail
+  }
+
+  weekly_snapshot = var.snapshot
+}
+
+variable "name" {
+  type    = string
+  default = "Report Rule 1"
```
afiune, line 36: report rule name is not unique?

examples/resource_lacework_report_rule/main.tf
```diff
@@ -0,0 +1,101 @@
...
+}
+
+variable "severities" {
+  type    = list(string)
+  default = ["High", "Medium"]
+}
+
+variable "resource_groups" {
+  type    = list(string)
+  default = ["TECHALLY_8416B4ADCED28565254842AA5906B729174653E1725F107"]
```
afiune, line 51: what happens if we accidentally remove this group, or how do we set this up for QA environments?

examples/resource_lacework_report_rule/main.tf
```diff
@@ -0,0 +1,101 @@
...
+}
+
+variable "channels" {
+  type    = list(string)
+  default = ["TECHALLY_2F0C086E17AB64BEC84F4A5FF8A3F068CF2CE15847BCBCA"]
```
afiune, line 56: yeah, this test is a bit fragile, we should rethink this.

website/docs/r/report_rule.html.markdown
```diff
@@ -0,0 +1,138 @@
...
+subcategory: "Report Rules"
+layout: "lacework"
+page_title: "Lacework: lacework_report_rule"
+description: |-
+  Create and manage Lacework Report Rules
+---
+
+# lacework\_report\_rule
+
+Use this resource to create a Lacework Report Rule in order to route reports to an email alert channel.
```
afiune, line 11:
```suggestion
Use this resource to create a Lacework Report Rule in order to route reports to one or more email alert channels.
```

website/docs/r/report_rule.html.markdown
```diff
@@ -0,0 +1,138 @@
...
+  name         = "GCP Resource Group"
+  description  = "All Gcp Projects"
+  organization = "MyGcpOrg"
+  projects     = ["*"]
+}
+
+resource "lacework_report_rule" "example" {
+  name             = "My Report Rule"
+  description      = "This is an example report rule"
+  channels         = [lacework_report_channel_slack.ops_critical.id]
```
afiune, line 53: ?? this seems wrong

website/docs/r/report_rule.html.markdown
```diff
@@ -0,0 +1,138 @@
...
+  }
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The report rule name.
+* `channels` - (Required) The list of report channels for the rule to use.
```
afiune, line 68: did you mean alert channels, also, they are email alert channels

website/docs/r/report_rule.html.markdown
```diff
@@ -0,0 +1,138 @@
...
+* `azure_compliance_reports` - (Optional) Compliance reports for Azure. See [Azure Compliance Reports](#Azure Compliance Reports) below for details.
+* `gcp_compliance_reports` - (Optional) Compliance reports for Gcp. See [Gcp Compliance Reports](#Gcp Compliance Reports) below for details.
+* `daily_compliance_reports` - (Optional) Daily event summary reports. See [Daily Compliance Reports](#Daily Compliance Reports) below for details.
+* `weekly_snapshot` - (Optional) A weekly compliance trend report for all monitored resources. Defaults to `false`.
+
+### Aws Compliance Reports
+
+`aws_compliance_reports` supports the following arguments:
+
+* `cis_s3` - (Optional) Defaults to `false`.
```
afiune, line 84: could you please add a long description in both places, here in the website documentation as well as every field inside the Schema? You can get the actual report names/sentences from the UI:
Screen Shot 2021-11-29 at 7 45 57 PM

lacework/resource_lacework_report_rule.go
```diff
@@ -0,0 +1,576 @@
...
+		return nil, fmt.Errorf(
+			"unable to import Lacework resource. Report Rule with guid '%s' was not found",
+			d.Id(),
+		)
+	}
+	log.Printf("[INFO] Report Rule found with guid: %s\n", response.Data.Guid)
+	return []*schema.ResourceData{d}, nil
+}
+
+func getReportRuleNotifications(d *schema.ResourceData) (api.ReportRuleNotifications, error) {
```
afiune, line 469: why do we return `(api.ReportRuleNotifications, error)` if there is not even one error being returned?

lacework/resource_lacework_report_rule.go
```diff
@@ -0,0 +1,576 @@
...
+
+	reportRule, err := api.NewReportRule(d.Get("name").(string),
+		api.ReportRuleConfig{
+			Description:        d.Get("description").(string),
+			Severities:         severities,
+			ResourceGroups:     castStringSlice(resourceGroups),
+			EmailAlertChannels: castStringSlice(channels),
+			NotificationTypes:  notifications,
+		},
+	)
```
afiune, line 408: Without the error check this could look much leaner like:
```suggestion
	reportRule, err := api.NewReportRule(d.Get("name").(string),
		api.ReportRuleConfig{
			Description:        d.Get("description").(string),
			Severities:         severities,
			ResourceGroups:     castStringSlice(resourceGroups),
			EmailAlertChannels: castStringSlice(channels),
			NotificationTypes:  getReportRuleNotifications(d),
		},
	)
```

lacework/resource_lacework_report_rule.go
```diff
@@ -0,0 +1,576 @@
...
+
+	reportRule.Guid = d.Id()
+
+	if !d.Get("enabled").(bool) {
+		reportRule.Filter.Enabled = 0
+	}
+
+	log.Printf("[INFO] Updating report rule with data:\n%+v\n", reportRule)
+	response, err := lacework.V2.ReportRules.Update(reportRule)
+
```
afiune, line 422: nit: it is better to not have empty lines in between running a function and checking if it returned an error
```suggestion
```

afiune (changes requested):

examples/resource_lacework_report_rule/main.tf
dmurray-lacework: It is not, but Report Rule 1 is not a good example name I will update this

examples/resource_lacework_report_rule/main.tf
dmurray-lacework: I will add in the tf inside the example to create and destroy this. Similar to what we did for alert rules

examples/resource_lacework_report_rule/main.tf
dmurray-lacework: Same as above i will add email alert channel to the terraform

website/docs/r/report_rule.html.markdown
dmurray-lacework: it should be email alert channels. Good catch!

website/docs/r/report_rule.html.markdown
dmurray-lacework: channels should be email_alert channel

website/docs/r/report_rule.html.markdown
dmurray-lacework: 👍

website/docs/r/report_rule.html.markdown
```diff
@@ -0,0 +1,137 @@
...
+  name         = "GCP Resource Group"
+  description  = "All Gcp Projects"
+  organization = "MyGcpOrg"
+  projects     = ["*"]
+}
+
+resource "lacework_report_rule" "example" {
+  name                 = "My Report Rule"
+  description          = "This is an example report rule"
+  email_alert_channels = [lacework_report_channel_slack.ops_critical.id]
```
afiune, line 53:
```suggestion
  email_alert_channels = [lacework_report_channel_email.team_email.id]
```

afiune (approved):
tenor-234880856

PR #247, opened 2021-12-14
fix: remove apiIntgKey from resourceLaceworkAlertChannelPagerDutyRead

dmurray-lacework:
***Description:***
PagerDuty Api has been updated to return apiIntgKey as empty string, this change removes apiIntgKey from resourceLaceworkAlertChannelPagerDutyRead and amends test assertions.


afiune (approved):
tenor-105144437

PR #275, opened 2022-02-25
docs: update query and policy resources

dmurray-lacework:
Update policy documentation to show a query resource being created and used for a policy resource.


afiune (approved):
Changed the majority of the docs since I synced up with @hazedav about the examples we were using.

PR #280, opened 2022-03-02
refactor: encourage the use of anonymous queries

afiune:
***Issue***: N/A

***Description:***
The motivation of this change it to avoid making our users set the query
id twice, it makes no sense to allow them to configure a different query
id, especially not since the platform enforces that both, the query id and
the query text match.

Furthermore, we have introduced the concept of anonymous queries, such
queries are encouraged and should the default way that users create and
manage LQL queries.

This change updates all documentation and tests to avoid adding the
query id inside the query text.

***Additional Info:***
Updated integration tests and documentation.

### DIFF
--- a/integration/resource_lacework_query_test.go
+++ b/integration/resource_lacework_query_test.go
@@ -120,8 +120,63 @@ func TestQueryCreate(t *testing.T) {
 	assert.Equal(t, queryStringK8, actualQuery)
 }
 
+func TestQueryDeprecatedSytaxWithID(t *testing.T) {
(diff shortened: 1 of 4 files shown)


integration/resource_lacework_query_test.go
```diff
@@ -176,4 +231,22 @@ var (
         EVENT
     }        
 }`
+
+	queryDeprecatedSyntaxWithID = `Lql_Terraform_Query {
```
afiune, line 235: I added one test to:
1. show that this method has been deprecated and
2. we still support it

PR #301, opened 2022-04-14
feat(resource): New lacework_alert_profile resource

dmurray-lacework:
***Issue***: https://lacework.atlassian.net/browse/ALLY-960

***Description:***
Add new resource lacework_alert_profile


integration/integration.go
```diff
@@ -227,6 +227,17 @@ func GetAlertRuleProps(result string) api.AlertRuleResponse {
 	return data
 }
 
+func GetAlertProfileProps(result string) api.AlertProfileResponse {
+	id := GetSpecificIDFromTerraResults(1, result)
+
+	var data api.AlertProfileResponse
+	err := LwClient.V2.Alert.Profiles.Get(id, &data)
+	if err != nil {
+		log.Fatalf("Unable to retrieve alert rule with id: %s", id)
```
dmurray-lacework, line 236:
```suggestion
		log.Fatalf("Unable to retrieve alert profile with id: %s", id)
```

integration/resource_lacework_alert_profile_test.go
```diff
@@ -0,0 +1,64 @@
...
+	actualDescription = terraform.Output(t, terraformOptions, "alert_description")
+
+	assert.Equal(t, "LW_CFG_GCP_DEFAULT_PROFILE", updateProps.Data.Extends)
+	assert.Equal(t, "{{_OCCURRENCE}} violation for GCP Resource {{RESOURCE_TYPE}}:{{RESOURCE_ID}} in project {{PROJECT_ID}} of organization {{ORGANIZATION}} region {{RESOURCE_REGION}} Updated", updateProps.Data.Alerts[0].Description)
+	assert.Equal(t, "Violation", updateProps.Data.Alerts[0].Name)
+	assert.Equal(t, "LW Configuration GCP Violation Alert", updateProps.Data.Alerts[0].EventName)
+	assert.Equal(t, "{{_OCCURRENCE}} violation detected in project {{PROJECT_ID}}", updateProps.Data.Alerts[0].Subject)
+
+	assert.Equal(t, "LW_CFG_GCP_DEFAULT_PROFILE", actualExtends)
+	assert.Equal(t, "LW_CFG_GCP_DEFAULT_PROFILE", actualExtends)
```
dmurray-lacework, line 62:
```suggestion
```

lacework/casting.go
```diff
@@ -108,6 +109,31 @@ func castAttributeToArrayOfCustomKeyValueMap(d *schema.ResourceData, attr string
 	return aMap
 }
 
+func castAttributeArrayToArrayOfType[T any](d *schema.ResourceData, attr string, typeList *[]T) (err error) {
+	var myType T
+	list := d.Get(attr).(*schema.Set).List()
```
dmurray-lacework, line 114:
```suggestion
	var (
		myType T
		list = d.Get(attr).(*schema.Set).List()
	)
```

lacework/resource_lacework_alert_profile_test.go
```diff
@@ -0,0 +1,26 @@
...
+		actual       = []api.AlertTemplate{{Name: "alertName", EventName: "alertEventName", Description: "alertDescription", Subject: "alertSubject"}}
+	)
+
+	_ = testResource.Set("alert", input)
+
+	err := castSchemaSetToArrayOfAlertTemplate(testResource, "alert", &alerts)
+	assert.NoError(t, err)
+	assert.Equal(t, alerts, actual,
+		"%s did not match expected value: %s", actual, alerts,
+	)
```
afiune, line 25: this test is really small, I feel we are not covering all cases with only one test like this, we should use a table test with multiple cases

lacework/resource_lacework_alert_profile.go
```diff
@@ -0,0 +1,219 @@
...
+		list []any
+	)
+
+	if d.Get(attr) == nil {
+		return errors.Errorf("attribute %s not found", attr)
+	}
+
+	list = d.Get(attr).(*schema.Set).List()
+	for _, item := range list {
+		val := sanitizeAlertTemplateKeys(item.(map[string]interface{}))
```
afiune, line 190: there is nothing that enforces that `item` IS actually a `map[string]interface{}` to this code could actually panic. I recommend doing this instead:
```suggestion
		iMap, ok := item.(map[string]interface{})
		if !ok {
			// maybe even print a log or something?
			continue
		}
		val := sanitizeAlertTemplateKeys(iMap)
```

lacework/resource_lacework_alert_profile_test.go
dmurray-lacework: The cast is now specific to alert profiles rather than generic, going to leaving the test as is.

lacework/resource_lacework_alert_profile.go
dmurray-lacework: 👍

lacework/schema_validators.go
```diff
@@ -22,3 +22,22 @@ func ValidSeverity() schema.SchemaValidateDiagFunc {
 		}
 	})
 }
+
+// StringDoesNotHavePrefix returns a SchemaValidateFunc which validates that the
+// provided value does not start with any of the chars.
+func StringDoesNotHavePrefix(chars string) schema.SchemaValidateDiagFunc {
```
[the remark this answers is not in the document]
dmurray-lacework: I'm not sure without mocking out the resource in a unit test. Easier to check in a Terratest

lacework/resource_lacework_alert_profile.go
```diff
@@ -0,0 +1,224 @@
...
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:        schema.TypeString,
+							Description: "The name that policies can use to refer to this template when generating alerts",
+							Required:    true,
+						},
+						"event_name": {
+							Type:        schema.TypeString,
+							Description: "The name of the resulting alert",
```
rmoles, line 51: nit
```suggestion
							Description: "The name of the resulting alert event",
```

lacework/resource_lacework_alert_profile.go
```diff
@@ -0,0 +1,224 @@
...
+							Required:    true,
+						},
+						"description": {
+							Type:        schema.TypeString,
+							Description: "The summary of the resulting alert",
```
rmoles, line 56: nit:
```suggestion
							Description: "The summary of the resulting alert event",
```

lacework/resource_lacework_alert_profile.go
```diff
@@ -0,0 +1,224 @@
...
+							Description: "A high-level observation of the resulting alert",
+							Required:    true,
+						},
+					},
+				},
+			},
+			"fields": {
+				Type:        schema.TypeSet,
+				Computed:    true,
+				Description: "A field is a declaration of a field to be mapped in from an LQL query.",
```
rmoles, line 70: I'm not sure I follow this description. Could this be re-worded?

rmoles: Also, a nitpick here. We are using `.` here and not in other descriptions

lacework/resource_lacework_alert_profile.go
```diff
@@ -0,0 +1,224 @@
...
+		alerts   []api.AlertTemplate
+	)
+	err := castSchemaSetToArrayOfAlertTemplate(d, "alert", &alerts)
+	if err != nil {
+		return err
+	}
+
+	alertProfile := api.NewAlertProfile(d.Get("name").(string),
+		d.Get("extends").(string),
+		alerts)
```
rmoles, line 89: Do we need any error checking here when using `Get` and subsequently casting to string?

dmurray-lacework: We don't, we use this in all our resources. The resource itself determines the type, we know it's always a string

lacework/resource_lacework_alert_profile.go
dmurray-lacework: I'm going to leave these as is, they're lifted from the api docs

lacework/resource_lacework_alert_profile.go
dmurray-lacework: I'm going to leave these as is, they're lifted from the api docs

website/docs/r/alert_profile.html.markdown
```diff
@@ -0,0 +1,51 @@
...
+subcategory: "Alert Profiles"
+layout: "lacework"
+page_title: "Lacework: lacework_alert_profile"
+description: |-
+  Create and manage Lacework Alert Profiles
+---
+
+# lacework\_alert\_profile
+
+Use this resource to create a Lacework Alert Profile in order map query results to events and form event descriptions.
```
rmoles, line 11:
```suggestion
Use this resource to create a Lacework Alert Profile in order to map query results to events and event descriptions.
```

rmoles, line 11:
```suggestion
Use this resource to create a Lacework Alert Profile in order to map query results to events and form event descriptions.
```

rmoles (approved):
Nice. (just one small comment to address) LGTM!

PR #316, opened 2022-06-01
feat: add expiry field for Vulnerability Exceptions

dmurray-lacework:

***Issue***: 
https://lacework.atlassian.net/browse/ALLY-1001

***Description:***
Add expiry field to vulnerability exception host and container resources.

### DIFF
--- a/lacework/resource_lacework_vulnerability_exception_container.go
+++ b/lacework/resource_lacework_vulnerability_exception_container.go
@@ -224,12 +224,11 @@ func resourceLaceworkVulnerabilityExceptionContainer() *schema.Resource {
 	}
 }
 
-func resourceLaceworkVulnerabilityExceptionContainerCreate(d *schema.ResourceData, meta interface{}) (err error) {
+func resourceLaceworkVulnerabilityExceptionContainerCreate(d *schema.ResourceData, meta interface{}) error {
 	var (
 		lacework   = meta.(*api.Client)
 		severities = castAttributeToStringSlice(d, "vulnerability_criteria.0.severities")
 		packages   = castAttributeToArrayOfCustomKeyValueMap(d, "vulnerability_criteria.0.package", "name", "version")
-		expiryTime time.Time
 		vulnExCfg  = api.VulnerabilityExceptionConfig{
 			Description:     d.Get("description").(string),
 			Type:            api.VulnerabilityExceptionTypeContainer,
@@ -242,12 +241,12 @@ func resourceLaceworkVulnerabilityExceptionContainerCreate(d *schema.ResourceDat
 	)
 
 	if d.Get("expiry") != nil {
-		expiryTime, err = time.Parse(time.RFC3339, d.Get("expiry").(string))
+		expiryTime, err := time.Parse(time.RFC3339, d.Get("expiry").(string))
 		if err != nil {
-			return errors.Wrap(err, fmt.Sprintf("unable to parse expiryTime %s", d.Get("expiry").(string)))
+			return errors.Wrap(err, fmt.Sprintf("unable to parse expiry time %s", d.Get("expiry").(string)))
 		}
+		vulnExCfg.ExpiryTime = expiryTime
 	}
-	vulnExCfg.ExpiryTime = expiryTime
 
 	if len(d.Get("resource_scope").([]interface{})) > 0 {
 		vulnExCfg.ResourceScope = api.VulnerabilityExceptionContainerResourceScope{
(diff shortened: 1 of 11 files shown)


lacework/schema_validators.go
```diff
@@ -41,3 +42,22 @@ func StringDoesNotHavePrefix(chars string) schema.SchemaValidateDiagFunc {
...
 }
+
+// ValidateTimeFormat returns a SchemaValidateFunc which validates that the
+// value is in the timeformat supplied.
+func ValidateTimeFormat(timeFormat string) schema.SchemaValidateDiagFunc {
+	return validation.ToDiagFunc(func(i interface{}, k string) (warnings []string, errors []error) {
+		v, ok := i.(string)
+		if !ok {
+			errors = append(errors, fmt.Errorf("expected type of %s to be string", k))
+			return warnings, errors
```
afiune, line 53: use naked returns instead
```suggestion
			return
```

lacework/schema_validators.go
```diff
@@ -41,3 +42,22 @@ func StringDoesNotHavePrefix(chars string) schema.SchemaValidateDiagFunc {
...
+		}
+
+		if _, err := time.Parse(timeFormat, v); err != nil {
+			errors = append(errors, fmt.Errorf("%s is not in format %s", v, timeFormat))
+			return warnings, errors
+		}
+
+		return warnings, errors
```
afiune, line 61: same, use naked returns instead
```suggestion
		}

		return
```

lacework/resource_lacework_vulnerability_exception_host.go
```diff
@@ -317,6 +335,14 @@ func resourceLaceworkVulnerabilityExceptionHostUpdate(d *schema.ResourceData, me
 		}
 	)
 
+	if d.Get("expiry") != nil {
+		expiryTime, err = time.Parse(time.RFC3339, d.Get("expiry").(string))
+		if err != nil {
+			return errors.Wrap(err, fmt.Sprintf("unable to parse expiryTime %s", d.Get("expiry").(string)))
```
afiune, line 341: where is `expiryTime` defined? 🤔 maybe this is better?
```suggestion
			return errors.Wrap(err, fmt.Sprintf("unable to parse expiry time %s", d.Get("expiry").(string)))
```

lacework/resource_lacework_vulnerability_exception_host.go
```diff
@@ -317,6 +335,14 @@ func resourceLaceworkVulnerabilityExceptionHostUpdate(d *schema.ResourceData, me
...
+		}
+	}
+	vulnExCfg.ExpiryTime = expiryTime
```
afiune, line 344: we don't need to "always" set this field, only if expiry time is set, so let us move this line to inside the if statement.

lacework/resource_lacework_vulnerability_exception_container.go
```diff
@@ -312,11 +330,12 @@ func resourceLaceworkVulnerabilityExceptionContainerRead(d *schema.ResourceData,
 	return nil
 }
 
-func resourceLaceworkVulnerabilityExceptionContainerUpdate(d *schema.ResourceData, meta interface{}) error {
+func resourceLaceworkVulnerabilityExceptionContainerUpdate(d *schema.ResourceData, meta interface{}) (err error) {
```
afiune, line 333: why do you use named returned here? 🤔

lacework/resource_lacework_vulnerability_exception_container.go
```diff
@@ -232,6 +241,14 @@ func resourceLaceworkVulnerabilityExceptionContainerCreate(d *schema.ResourceDat
 		}
 	)
 
+	if d.Get("expiry") != nil {
+		expiryTime, err = time.Parse(time.RFC3339, d.Get("expiry").(string))
+		if err != nil {
+			return errors.Wrap(err, fmt.Sprintf("unable to parse expiryTime %s", d.Get("expiry").(string)))
+		}
+	}
+	vulnExCfg.ExpiryTime = expiryTime
```
afiune, line 250: same, we don't use `expiryTime` anywhere else in this function (and other ones), move it to inside the if statement

afiune (changes requested):

lacework/resource_lacework_vulnerability_exception_container.go
```diff
@@ -402,7 +427,10 @@ func vulnerabilityExceptionReasonsSlice() []string {
 }
 
 func vulnerabilityExceptionFixableEnabled(fixable []int) bool {
```
afiune, line 429: This code is in both, the go-sdk and the provider, we should probably move it to the go-sdk and consume it from there. https://lacework.atlassian.net/browse/ALLY-1033

afiune (approved):

PR #327, opened 2022-06-17
feat(GcpGkeAudit): Add support for GcpGkeAudit apiv2 cloud account integration

rmoles:
***Issue***: Include link to the Jira/Github Issue
https://lacework.atlassian.net/browse/ALLY-949

***Description:***
In order to facilitate the creation of the GCP GKE Audit Log Terraform Module, we need to add support for the GcpGkeAudit cloud account integration

***Additional Info:***
Currently blocked from releasing the go-sdk implementation for `GcpGkeAudit` so unable to run the tests successfully


lacework/resource_lacework_integration_gcp_gke_audit_log.go
```diff
@@ -0,0 +1,305 @@
...
+				return nil, []error{
+					fmt.Errorf("%s: can only be either 'PROJECT' or 'ORGANIZATION'", key),
+				}
+			}
+		},
+	},
+	"organization_id": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "The GCP Organization ID (required when integration_type is organization).",
```
afiune, line 117: If this field is required when the integration type is `ORGANIZATION`, could we add a validation that checks it and returns a helpful message to the end-user?

afiune (changes requested):
One comment, plus the fact that you haven't add any documentation.

Out of that, code looks good to go.

lacework/resource_lacework_integration_gcp_gke_audit_log.go
rmoles: @afiune. Do we want to catch this at this level or should we let the API return an error on this? Generally I see Terraform doing the latter and letting the API do the validation.

afiune: I am ok leaving it to the API, though, we try to add validation in resources so that the user detects these kinds of things early on (that is, during `terraform plan`) rather than in the middle of an apply.

Up to you to decide what's best for our users here.

rmoles: Sure I'll see what I can do in terms of validation at this level. Looks like `CustomizeDiff` might give us what we need to validate this.

PR #328, opened 2022-06-17
test(policy): fix policy type in integration testing

hazedav:
***Issue***:
https://lacework.atlassian.net/browse/ALLY-1050

***Description:***
Invalid policy type in integration testing for policy resource

### DIFF
--- a/integration/resource_lacework_policy_test.go
+++ b/integration/resource_lacework_policy_test.go
@@ -63,7 +63,6 @@ func TestPolicyCreate(t *testing.T) {
 	terraformOptions.Vars = map[string]interface{}{
 		"title":       "lql-terraform-policy-updated",
 		"severity":    "Low",
-		"type":        "Compliance",
 		"description": "Policy Created via Terraform Updated",
 		"remediation": "Please Ignore",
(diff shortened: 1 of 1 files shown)


integration/resource_lacework_policy_test.go
```diff
@@ -63,7 +63,7 @@ func TestPolicyCreate(t *testing.T) {
 	terraformOptions.Vars = map[string]interface{}{
 		"title":       "lql-terraform-policy-updated",
 		"severity":    "Low",
-		"type":        "Summary",
+		"type":        "Compliance",
```
varming, line 66: "Violation" is the only policy type accepted by the API, so this will get a failure (I hope). My guess is that you can remove this field from the update and you have a working test.

hazedav: 👍

dkoshe (approved):

hazedav:
@dmurray-lacework , It looks like we're not tearing these policies down as part of the test framework.  Is this something we have the capabilities for?

afiune (approved):

PR #331, opened 2022-06-21
fix: handle resource deleted outside of terraform scope

dmurray-lacework:

***Issue***: 
https://lacework.atlassian.net/browse/ALLY-1053

### DIFF
--- a/lacework/errors.go
+++ b/lacework/errors.go
@@ -8,7 +8,7 @@ import (
 )
 
 func notFound(err error) bool {
-	return strings.Contains(err.Error(), "404")
+	return strings.Contains(err.Error(), "[404]")
 }
 
 func resourceNotFound(d *schema.ResourceData, err error) error {
--- a/lacework/errors.go
+++ b/lacework/errors.go
@@ -1,16 +1,14 @@
 package lacework
 
 import (
-	"errors"
 	"log"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


lacework/errors.go
```diff
@@ -1,16 +1,14 @@
...
+	"strings"
 
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
 func notFound(err error) bool {
-	var e *resource.NotFoundError
-	return errors.As(err, &e)
+	return strings.Contains(err.Error(), "404")
```
rmoles, line 11: When I look at the error in the failing pipeline in customer demo I see the error is
```
[GET] https://customerdemo.lacework.net/api/v2/VulnerabilityExceptions/VULN_A47591B615EC330E0CDDEE28D177FDE5A937FC62553EE18B587                                                                                                                                                              
[404] Not found   
```
Will `err.Error()` be this entire error string above? If so, is there any danger that the VULN GUID could contain `404`?

rmoles: Not sure if it comes through exactly as above, but perhaps we could check that it contains `[404]`?

afiune (approved):
Could we please write a quick test? What if the error message changes?

Create something and then delete it with the go api client, then test this functionality, please.

PR #332, opened 2022-06-23
fix: time parse error for expiry field

dmurray-lacework:
***Issue***:  https://lacework.atlassian.net/browse/ALLY-1055

***Description:***
- Check for empty string before attempting to parse expiry
- Add  integration test with no expiry field

### DIFF
--- a/integration/resource_lacework_vulnerability_exception_container_test.go
+++ b/integration/resource_lacework_vulnerability_exception_container_test.go
@@ -42,6 +42,9 @@ func TestVulnerabilityExceptionContainerCreate(t *testing.T) {
 	assert.Equal(t, "[CVE-2016-9840 CVE-2018-14599 CVE-2018-6942]", actualCves)
 	assert.Equal(t, "[map[name:adm-zip version:0.4.7] map[name:myPackage version:1.0.0]]", actualPackages)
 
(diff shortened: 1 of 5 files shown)


integration/resource_lacework_vulnerability_exception_container_test.go
```diff
@@ -42,6 +42,9 @@ func TestVulnerabilityExceptionContainerCreate(t *testing.T) {
 	assert.Equal(t, "[CVE-2016-9840 CVE-2018-14599 CVE-2018-6942]", actualCves)
 	assert.Equal(t, "[map[name:adm-zip version:0.4.7] map[name:myPackage version:1.0.0]]", actualPackages)
 
+	noexpiry := GetSpecificIDFromTerraResults(2, create)
```
rmoles, line 45: If I'm understanding this correctly, we are trying to pull out the 2nd resource from the terraform result by index?
Should this be this instead?
```suggestion
	noexpiry := GetSpecificIDFromTerraResults(1, create)
```

dmurray-lacework: No, the input is the length of the resultant matched id's not the index.

rmoles (approved):
Awesome. LGTM!

Issue #339, opened 2022-07-08
fix: vulnerability exception resources should not always pass Fixable field


matthewashley1:
# Feature Request

**Describe the Feature Request**
To have the fixable parameter of vulnerability criteria for a vulnerability exception container to not get defined by default.

**Is your feature request related to a problem? Please describe**
I am attempting the setup vulnerability exceptions and would like to set exceptions for multiple CVES. I however do not care if a CVE is fixable or not and would not like to constrain the scope of my exceptions for fixable. When creating vulnerability exceptions with the lacework terraform provider the fixable parameter gets set by default to false causing my inline scanner runs in my jenkins pipeline to fail. I want to define one vulnerability exception for multiple CVES that are a mix of having a fix version or not. Currently, the lacework terraform provider is not allowing me to do that as the fixable parameter is set by default. 

**Describe Preferred Solution**
To have the ability to not set the fixable parameter of vulnerability criteria for a vulnerability exception container.

**Additional Context**
Provider Version: 0.20.1

afiune:
Hi @matthewashley1 - Thank you so much for your feedback!

I looked quickly in the code and it indeed looks we are always passing the Fixable parameter, I have created an
internal ticket for my team to tackle this issue in our next sprint, we will let you know once this is fixed.

- https://lacework.atlassian.net/browse/ALLY-1085

afiune:
Hi @matthewashley1 👋🏽 - We closed this issue before making the modifications in our Terraform provider.

![tenor-63466038](https://user-images.githubusercontent.com/5712253/182282045-811d8ccc-5d27-4a29-8556-76fae14d78fa.gif)

Good news, we just fixed this issue! 🙌🏽

![tenor-143493384](https://user-images.githubusercontent.com/5712253/182282169-64b9c830-9fec-42f4-ac88-37b736119f29.gif)

Unfortunately, Terraform doesn't have (yet) a great solution for this cases where we need to have a tristate
boolean. 

- Read more about this here -> https://github.com/hashicorp/terraform-plugin-sdk/issues/817

The approach we took was to introduce a new argument inside the `vulnerability_criteria` named `fixable_vuln`,
this new argument is of type `string` and it solves the tristate boolean issue.

The (tres) states are:
* `"false"` (Yes! A string with value `false`) - Constraints the exception by **non-fixable** vulnerabilities
* `"true"` (A string with value `true`) - Constraints the exception by **fixable** vulnerabilities
[...]

PR #342, opened 2022-07-14
fix: fixable parameter for vulnerability_exception resources

dmurray-lacework:

***Issue***: 
https://lacework.atlassian.net/browse/ALLY-1085
Fixes https://github.com/lacework/terraform-provider-lacework/issues/339

***Description:***
Allow fixable to be nil value


integration/integration.go
```diff
@@ -231,8 +231,8 @@ func GetAlertChannelProps(result string) api.AlertChannelResponse {
 	return data
 }
 
-func GetVulnerabilityExceptionProps(result string) api.VulnerabilityExceptionResponse {
-	id := GetSpecificIDFromTerraResults(1, result)
+func GetVulnerabilityExceptionProps(result string, resourceID int) api.VulnerabilityExceptionResponse {
```
afiune, line 234: can we roll back this?

integration/resource_lacework_vulnerability_exception_container_test.go
```diff
@@ -66,3 +65,44 @@ func TestVulnerabilityExceptionContainerCreate(t *testing.T) {
...
+		"description":     "Updated Vulnerability Exception created by Terraform",
+		"cves":            []string{"CVE-2016-9840", "CVE-2018-6940"},
+		"package_name":    "myUpdatedPackage",
+		"package_version": "1.1.0",
+	}
+
+	update := terraform.ApplyAndIdempotent(t, terraformOptions)
+	updateProps := GetVulnerabilityExceptionProps(update, 1)
+
+	assert.Nil(t, updateProps.Data.VulnerabilityCriteria.FixableEnabled())
```
afiune, line 107: We need to add two tests (which means two Terraform applies)

- One is to SET the fixable field to either `true` or `false` (not nil)
- The second one to revert the change and UNSET the fixable field (put it back to nil)

afiune (approved):

lacework/resource_lacework_vulnerability_exception_container.go
```diff
@@ -123,10 +124,18 @@ func resourceLaceworkVulnerabilityExceptionContainer() *schema.Resource {
 								},
 							},
 						},
+						"fixable_vuln": {
+							Type:         schema.TypeString,
+							Optional:     true,
+							Description:  "The state of the vulnerability exception",
```
afiune, line 130: update this description

lacework/resource_lacework_vulnerability_exception_container.go
```diff
@@ -123,10 +124,18 @@ func resourceLaceworkVulnerabilityExceptionContainer() *schema.Resource {
...
+							ValidateFunc: validation.StringInSlice([]string{"true", "false"}, true),
+						},
 						"fixable": {
-							Type:        schema.TypeBool,
-							Optional:    true,
-							Description: "The state of the vulnerability exception",
+							Type:          schema.TypeBool,
+							Optional:      true,
+							Description:   "The state of the vulnerability exception",
+							Deprecated:    "This attribute deprecated and has been replaced by `fixable_vuln`",
```
afiune, line 137:
```suggestion
							Deprecated:    "This attribute is deprecated and has been replaced by `fixable_vuln`",
```

lacework/resource_lacework_vulnerability_exception_container.go
```diff
@@ -229,17 +238,30 @@ func resourceLaceworkVulnerabilityExceptionContainerCreate(d *schema.ResourceDat
...
 	)
 
+	// leave deprecated field 'fixable' behaviour
+	if _, ok := d.GetOk("vulnerability_criteria.0.fixable"); ok {
+		fixablePtr := d.Get("vulnerability_criteria.0.fixable").(bool)
+		fixable = &fixablePtr
+	}
+
+	if _, ok := d.GetOk("vulnerability_criteria.0.fixable_vuln"); ok {
+		fixablePtr, _ := strconv.ParseBool(d.Get("vulnerability_criteria.0.fixable_vuln").(string))
```
afiune, line 251: please check the error

lacework/resource_lacework_vulnerability_exception_container.go
```diff
@@ -322,7 +344,15 @@ func resourceLaceworkVulnerabilityExceptionContainerRead(d *schema.ResourceData,
 	vulnCriteria["severities"] = response.Data.VulnerabilityCriteria.Severity
 	vulnCriteria["cves"] = response.Data.VulnerabilityCriteria.Cve
 	vulnCriteria["package"] = vulnerabilityExceptionFlattenPackages(response.Data.VulnerabilityCriteria.Package)
-	vulnCriteria["fixable"] = response.Data.VulnerabilityCriteria.FixableEnabled()
+
+	if response.Data.VulnerabilityCriteria.FixableEnabled() != nil {
+		if _, ok := d.GetOk("vulnerability_criteria.0.fixable"); ok {
+			vulnCriteria["fixable"] = response.Data.VulnerabilityCriteria.FixableEnabled()
+		} else {
+			vulnCriteria["fixable_vuln"] = response.Data.VulnerabilityCriteria.FixableEnabled()
```
afiune, line 352: This should be converted to a string? I mean, `FixableEnabled()` returns a boolean pointer so 🤔

lacework/resource_lacework_vulnerability_exception_host.go
```diff
@@ -218,17 +218,19 @@ func resourceLaceworkVulnerabilityExceptionHostCreate(d *schema.ResourceData, me
 	var (
 		lacework   = meta.(*api.Client)
 		severities = castAttributeToStringSlice(d, "vulnerability_criteria.0.severities")
+		fixable    = d.Get("vulnerability_criteria.0.fixable").(bool)
```
afiune, line 221: this looks different from container. can we fix host and do what you did in container?

website/docs/r/vulnerability_exception_container.html.markdown
```diff
@@ -64,7 +64,7 @@ The following arguments are supported:
   `Critical`, `High`, `Medium`, `Low` and `Info`.
 * `cves` - (Optional) The list of the cves.
 * `package` - (Optional) The list of the packages.
-* `fixable` - (Optional) Whether to filter on fixable. Defaults to `false`.
+* `fixable_vuln` - (Optional) Whether to filter on fixable.
```
afiune, line 67:
```suggestion
* `fixable_vuln` - (Optional) Whether to filter on fixable. This argument is of type `string` and the valid values are: `"true"`, `"false"`, and `""`. An empty string represents the absence of the filter.
```

afiune: update this on host too, please

integration/resource_lacework_vulnerability_exception_container_test.go
```diff
@@ -65,4 +65,61 @@ func TestVulnerabilityExceptionContainerCreate(t *testing.T) {
 	assert.Equal(t, "Updated Vulnerability Exception created by Terraform", actualDescription)
 	assert.Equal(t, "[CVE-2016-9840 CVE-2018-6940]", actualCves)
 	assert.Equal(t, "[map[name:adm-zip version:0.4.7] map[name:myUpdatedPackage version:1.1.0]]", actualPackages)
```
afiune, line 67: can we check if fixable is true?

examples/resource_lacework_vulnerability_exception_container/fixable/main.tf
```diff
@@ -78,4 +84,8 @@ output "cves" {
 
 output "packages" {
   value = lacework_vulnerability_exception_container.example.vulnerability_criteria.0.package
+}
+
+output "fixable" {
+  value = lacework_vulnerability_exception_container.example.vulnerability_criteria.0.fixable
```
afiune, line 90: The problem with your drift wasn't in the provider codebase, but instead here in your test.

You can use this code to store the output as `"null"` when the value of `fixable` is `null` 
```suggestion
  value = coalesce(lacework_vulnerability_exception_container.example.vulnerability_criteria.0.fixable, "null")
```

PR #344, opened 2022-07-19
feat: support authentication via API token

afiune:
***Issue***: https://lacework.atlassian.net/browse/ALLY-1083

***Description:***
As a Lacework Terraform user,
I would like to configure the Terraform provider for Lacework with my account name and API token,
So I can re-use a token and not hit API Rame Limits

***Additional Info:***
For short-lived workflows, Lacework provides an alternative method to configure the provider by using an
API access token (`api_token`, `LW_API_TOKEN`), note that all API tokens from the Lacework platform are
short-lived which means that once the token expires, a new one needs to be generated and configured.

```hcl
provider "lacework" {
  account   = "my-account"
  api_token = "my-api-token"
}
```

**Using environment variables**

**Bash:**
```
export LW_ACCOUNT="my-account"
export LW_API_TOKEN="my-api-token"
```

**Powershell:**
```
$Env:LW_ACCOUNT = "my-account"
$Env:LW_API_TOKEN = "my-api-token"
```

**Note:** You can use the [Lacework CLI](https://docs.lacework.com/cli) command `lacework access-token` to
generate an API access token and the command `lacework configure show account` to display your configured account.

### DIFF
--- a/lacework/provider.go
+++ b/lacework/provider.go
@@ -51,6 +51,12 @@ func Provider() *schema.Provider {
 				DefaultFunc: schema.EnvDefaultFunc("LW_API_SECRET", nil),
 				Description: "Lacework API access secret",
 			},
+			"api_token": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("LW_API_TOKEN", nil),
+				Description: "Lacework API access token",
+			},
 			"organization": {
 				Type:        schema.TypeBool,
 				Optional:    true,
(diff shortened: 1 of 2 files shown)


hazedav:
Looks good to me!

lacework/provider.go
```diff
@@ -162,8 +169,16 @@ func providerConfigure(_ context.Context, d *schema.ResourceData) (interface{},
 		}
 	}
 
-	if account != "" && key != "" && secret != "" {
-		apiOpts = append(apiOpts, api.WithApiKeys(key, secret))
+	// authentication via environment variables or static credentials
+	if validStaticCredentials(account, key, secret, token) {
+		if token != "" {
+			apiOpts = append(apiOpts, api.WithToken(token))
```
mjunglw, line 175: Noob question. If token is `nil` does it still call `api.WithToken(token)`? I assume that logic is handled somewhere else and this codepath doesn't have to handle that logic.

afiune: You assumption is correct, Terraform has out-of-the-box schema validation, so every argument (token, key, account, etc.) has already been validated safely. The line you are looking at for the `token` string is:
```go
139	token = d.Get("api_token").(string)
```

rmoles (approved):
image

dmurray-lacework (approved):

PR #369, opened 2022-09-20
refactor: migrate gcp_at to use v2 api

dmurray-lacework:

***Issue***: 
https://lacework.atlassian.net/browse/ALLY-1139

***Description:***
Refactor gcp_at resource to use v2 api

### DIFF
--- a/lacework/resource_lacework_integration_gcp_at.go
+++ b/lacework/resource_lacework_integration_gcp_at.go
@@ -182,10 +182,6 @@ func resourceLaceworkIntegrationGcpAtCreate(d *schema.ResourceData, meta interfa
 			))
 		}
 
-		if err != nil {
-			return resource.NonRetryableError(err)
-		}
-
 		integration := response.Data
 		d.SetId(integration.IntgGuid)
 		d.Set("name", integration.Name)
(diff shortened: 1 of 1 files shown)


lacework/resource_lacework_integration_gcp_at.go
```diff
@@ -162,48 +163,41 @@ func resourceLaceworkIntegrationGcpAtCreate(d *schema.ResourceData, meta interfa
...
-				api.GcpAuditLogIntegration.String(), retries,
+				"unable to create %s integration (retrying %d more time(s))",
+				api.GcpAtSesCloudAccount.String(), retries,
 			))
 		}
 
-		log.Printf("[INFO] Verifying server response")
-		err = validateGcpIntegrationResponse(&response)
 		if err != nil {
 			return resource.NonRetryableError(err)
```
rmoles, line 186: Wondering if this could be moved inside the above if statement?

dmurray-lacework: it can be removed entirely, there err it was checking was from the func that was removed

dmurray-lacework: Fixed

PR #382, opened 2022-09-29
BREAKING 💥 Migrate resources to use V2 api

dmurray-lacework:
Migrate TF provider resources to use V2 of the api.


afiune:
Moved to draft to avoid approving and merging without Major version bump.

tenor-213370760

afiune (approved):
tenor-109701152

PR #385, opened 2022-10-13
feat(resource): New lacework_integration_aws_org_agentless_scanning resource

theopolis:
***Issue***: RAIN-38397

This replaces PR#384.

***Description:***
This adds the AWS Organizations Agentless Scanning type. This type is similar to the existing Agentless Scanning integration, but this allows a single scanning account to exist with multiple monitored accounts. The monitored accounts can be a mixed set of OUs, the organization root, or specific Account IDs.

The `lacework/terraform-aws-agentless-scanning` module will be used to install this type of account.


lacework/resource_lacework_integration_aws_org_agentless_scanning.go
```diff
@@ -0,0 +1,321 @@
...
+	)
+
+	awsOrgAgentlessScanningData := api.AwsSidekickOrgData{
+		ScanFrequency:           d.Get("scan_frequency").(int),
+		ScanContainers:          d.Get("scan_containers").(bool),
+		ScanHostVulnerabilities: d.Get("scan_host_vulnerabilities").(bool),
+		AccountID:               d.Get("account_id").(string),
+		BucketArn:               d.Get("bucket_arn").(string),
+		ScanningAccount:         d.Get("scanning_account").(string),
+		MonitoredAccounts:       strings.Join(d.Get("monitored_accounts").([]string), ", "),
```
dmurray-lacework, line 160: This cast fails,  we can call the following instead
```suggestion
		MonitoredAccounts:       strings.Join(castAttributeToStringSlice(d, "monitored_accounts"), ", "),
```

theopolis: Good call, TBH I was worried about this, as the proper way is to first try `[]interface{}` then for `string` on each element.

dmurray-lacework (approved):

PR #389, opened 2022-10-19
fix: suppressdiff for gcp_at private_key_id

dmurray-lacework:

***Issue***:  https://lacework.atlassian.net/browse/ALLY-1238

***Description:***
From v2 of the api the private key id field is a sensitive value and not returned. The diff needs to be suppressed for this field, otherwise there will be drift


afiune (approved):
tenor-30270670

PR #402, opened 2022-11-09
fix: azure_cfg resource sening incorrect cloud account type

dmurray-lacework:

Azure_cfg is sending the wrong cloud account type


rmoles (approved):
Good catch

PR #403, opened 2022-11-09
fix: azure_cfg resource sending incorrect cloud account type

dmurray-lacework:
azure_cfg resource is sending incorrect cloud account type


afiune (approved):
tenor-184070975

PR #412, opened 2022-11-16
test: change alert_profile test to read second element from alerts array

dmurray-lacework:

***Description:***
Fix alert profile test. Find alertTemplate assertion from returned alerts by name

### DIFF
--- a/integration/resource_lacework_alert_profile_test.go
+++ b/integration/resource_lacework_alert_profile_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/gruntwork-io/terratest/modules/terraform"
+	"github.com/lacework/go-sdk/api"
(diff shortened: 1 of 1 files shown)


integration/resource_lacework_alert_profile_test.go
```diff
@@ -32,10 +32,10 @@ func TestAlertProfileCreate(t *testing.T) {
 	actualDescription := terraform.Output(t, terraformOptions, "alert_description")
 
 	assert.Equal(t, "LW_CFG_GCP_DEFAULT_PROFILE", createProps.Data.Extends)
```
varming, line 34: I would change this test to find the alert by name. There will always be an alert named `Violation`, you can assert that. This way the test will not fail if a new alert is added.

afiune (approved):

PR #427, opened 2022-12-07
fix: add server token and uri computed fields to lacework_gcp_agentless_scanning

ammarekbote:
Issue: RAIN-43475

Description:
This fix exposes server_token for GcpSidekick


theopolis (approved):
I reviewed the changes to the GCP agentless scanner, looks good.

lacework/provider.go
```diff
@@ -326,14 +326,14 @@ func verifyPrimaryAccount(account string, opts ...api.Option) (string, error) {
...
 	}
 
-	orgInfo, err := lwApi.Account.GetOrganizationInfo()
+	orgInfo, err := lwApi.V2.OrganizationInfo.Get()
 	if err != nil {
 		return account, err
 	}
 
-	if orgInfo.OrgAccount {
+	if orgInfo.Data[0].OrgAccount {
```
afiune, line 334: could we please check `len(orgInfo.Data) > 0` before accessing `Data[0]` - otherwise the provider will panic.

lacework/resource_lacework_integration_gcp_agentless_scanning.go
```diff
@@ -265,6 +273,8 @@ func resourceLaceworkIntegrationGcpAgentlessScanningCreate(d *schema.ResourceDat
 		d.Set("created_or_updated_by", integration.CreatedOrUpdatedBy)
 		d.Set("type_name", integration.Type)
 		d.Set("org_level", integration.IsOrg == 1)
+		d.Set("server_token", integration.ServerToken)
+		d.Set("uri", integration.Uri)
```
afiune, line 277: there two `Set()`s need to be added to `resourceLaceworkIntegrationGcpAgentlessScanningRead()` and probably also to `resourceLaceworkIntegrationGcpAgentlessScanningUpdate()` - just to be safe.

afiune (changes requested):

lacework/resource_lacework_integration_gcp_agentless_scanning.go
afiune: Talking with @ammarekbote over Slack, he mentioned that the token won't be returned on GET and PATCH requests for security reasons, therefore adding the server token only in POST requests (create) is the right thing to do, having said that, the URI is shown all the time so let's just add URI on Read/Update functions here.

lacework/provider.go
ammarekbote: Done. Adding this check in next commit.

lacework/resource_lacework_integration_gcp_agentless_scanning.go
ammarekbote: I see that we populate both these fields for `update` too - so setting both there and for `read` just setting Uri

afiune (approved):

PR #429, opened 2022-12-08
feat(jira): add bidirectional configuration 

hazedav:
***Issue***:
https://lacework.atlassian.net/browse/ALLY-1220

***Description:***
This adds the bidirectional `configuration` option to the Lacework terraform provider


hazedav:
I'm 💯 percent certain that I'm missing something here (i.e. it seems like the tests require environment variables, but I don't see them being set anywhere in source code)...

afiune (approved):

PR #461, opened 2023-03-23
fix: avoid sending an empty GCP private_key_id

afiune:
***Issue***: https://lacework.zendesk.com/agent/tickets/11161

***Description:***
When doing a `terraform apply` without changing the private key or private key id, we send a `PATCH` request with empty fields, this is causing the following error message:

```
[400] Error validating integration. Please verify the configuration details entered.
```

This change avoids sending an empty GCP Private Key ID and it also avoids overriding empty fields doing `Read()` operations.
***Additional Info:***
Related to https://github.com/lacework/go-sdk/pull/1202


ipcrm (approved):
:shipit: LGTM!

PR #467, opened 2023-04-12
feat: Addition of S3 Bucket ARN for AWS EKS Audit log

djmctavish:
***Issue***: Include link to the Jira/Github Issue
https://lacework.atlassian.net/browse/GROW-1520

***Description:***
Added support for S3 Bucket ARN for the AWS EKS Audit log integration

***Additional Info:***


integration/resource_lacework_integration_aws_eks_audit_log_test.go
```diff
@@ -17,9 +17,10 @@ func TestIntegrationAwsEksAuditLog(t *testing.T) {
 		TerraformDir: "../examples/resource_lacework_integration_aws_eks_audit_log",
 		EnvVars:      tokenEnvVar,
 		Vars: map[string]interface{}{
-			"role_arn":    "arn:aws:iam::249446771485:role/lacework-iam-example-role",
-			"external_id": "12345",
-			"sns_arn":     "arn:aws:sns:us-west-2:123456789123:foo-lacework-eks",
+			"role_arn":      "arn:aws:iam::249446771485:role/lacework-iam-example-role",
+			"external_id":   "12345",
+			"sns_arn":       "arn:aws:sns:us-west-2:123456789123:foo-lacework-eks",
+			"s3_bucket_arn": "arn:aws:s3:::example-bucket-name",
```
jon-stewart, line 23: Does this suggest that our TF has no ARN validation?

jon-stewart (approved):

integration/resource_lacework_integration_aws_eks_audit_log_test.go
dmurray-lacework, line 23: Think there might be an issue when the new optional field is not supplied.

We might want to add another integration test without this field to confirm

dmurray-lacework (changes requested):
Getting a 400 when I remove the s3_bucket_arn from the example main.tf

If the new field is optional I think you may have to make a change to go-sdk to omit from json payload when the field is empty.
https://github.com/lacework/go-sdk/blob/main/api/cloud_accounts_aws_eks_audit.go#L51

integration/resource_lacework_integration_aws_eks_audit_log_test.go
djmctavish: Good point

examples/resource_lacework_integration_aws_eks_audit_log/audit_log_without_s3_arn/main.tf
```diff
@@ -0,0 +1,58 @@
...
+variable "sns_arn" {
+  type      = string
+  default   = "arn:aws:sns:us-west-2:123456789123:foo-lacework-eks"
+}
+
+variable "s3_bucket_arn" {
+  type      = string
+  default   = "arn:aws:s3:::example-bucket-name"
+}
+
```
dmurray-lacework, line 23: Can remove this variable from the example.

djmctavish: 👍

dmurray-lacework (approved):

PR #473, opened 2023-04-25
feat: Add org account mappings to Agentless for AWS

wl-smith:
***Issue***: (https://lacework.atlassian.net/browse/RAIN-54797)

***Description:***
This is step #2 to provide LW Org support for Terraform integrations for Agentless. Step 1 having been to update go-sdk: https://github.com/lacework/go-sdk/pull/1251. 

***Additional Info:***
The changes made were essentially to add the account mapping file to the awsOrgAgentlessScanningIntegrationSchema  in terraform, add helper functions to get & manipulate the account mapping input from the customer.

***Tests:***
Added a new integration test specifically for AwsSidekickOrg Integrations as well as updated the example terraform and the response object to get the AwsAgentlessOrgScanningResponse, which is a new function in integration.go that uses the LwOrgClient instead of the LwClient.


examples/resource_lacework_integration_aws_org_agentless_scanning/main.tf
```diff
@@ -6,6 +6,11 @@ terraform {
   }
 }
 
+provider "lacework" {
+  alias = "organization"
```
afiune, line 10: This might be the problem, if you configure an alias to this provider, you have to use it on every resource, an example is:
```hcl
resource "lacework_integration_aws_org_agentless_scanning" "example" {
    provider = lacework.organization
    name = "foo"
    ...
}
```

afiune (commented):
Maybe we can reuse most of the code at https://github.com/lacework/terraform-provider-lacework/blob/main/lacework/resource_lacework_integration_aws_ct.go#L318-L385?

afiune (approved):

afiune:
CI failure is NOT related to this change.

PR #492, opened 2023-06-01
feat(resource): add support for compliance policies

PengyuanZhao:
***Issue***: https://lacework.atlassian.net/browse/GROW-1659

***Description:***
Add new resource `resource_lacework_policy_compliance` to support the `Compliance` policy type

***Additional Info:***
Run `make integration-test regex=TestPolicyComplianceCreate` to test.

### DIFF
--- a/lacework/resource_lacework_policy_compliance.go
+++ b/lacework/resource_lacework_policy_compliance.go
@@ -10,8 +10,6 @@ import (
 	"github.com/lacework/go-sdk/api"
 )
 
-const policyType = "Compliance"
-
 func resourceLaceworkPolicyCompliance() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceLaceworkPolicyComplianceCreate,
@@ -73,21 +71,11 @@ func resourceLaceworkPolicyCompliance() *schema.Resource {
 					Type: schema.TypeString,
 				},
 			},
-			"alerting": {
-				Type:        schema.TypeList,
-				MaxItems:    1,
+			"alerting_enabled": {
+				Type:        schema.TypeBool,
 				Optional:    true,
-				Description: "Alerting",
-				Elem: &schema.Resource{
-					Schema: map[string]*schema.Schema{
-						"enabled": {
-							Type:        schema.TypeBool,
-							Optional:    true,
-							Default:     true,
-							Description: "Whether alerting is enabled or disabled",
-						},
-					},
-				},
+				Default:     true,
+				Description: "Whether alerting is enabled or disabled",
 			},
 			"id": {
 				Type:     schema.TypeString,
@@ -120,16 +108,16 @@ func resourceLaceworkPolicyComplianceCreate(d *schema.ResourceData, meta interfa
 	)
 
 	policy := api.NewPolicy{
-		PolicyType:   policyType,
+		PolicyType:   api.PolicyTypeCompliance.String(),
 		QueryID:      d.Get("query_id").(string),
 		Title:        d.Get("title").(string),
 		Enabled:      d.Get("enabled").(bool),
 		Description:  d.Get("description").(string),
 		Remediation:  d.Get("remediation").(string),
 		Severity:     d.Get("severity").(string),
-		AlertEnabled: d.Get("alerting.0.enabled").(bool),
 		PolicyID:     d.Get("policy_id_suffix").(string),
 		Tags:         castStringSlice(d.Get("tags").(*schema.Set).List()),
+		AlertEnabled: d.Get("alerting_enabled").(bool),
 	}
 
 	log.Printf("[INFO] Creating Policy with data:\n%+v\n", policy)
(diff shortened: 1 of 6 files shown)


lacework/resource_lacework_policy_compliance.go
```diff
@@ -0,0 +1,264 @@
...
+						return nil, nil
+					default:
+						return nil, []error{
+							fmt.Errorf(
+								"%s: can only be 'Compliance'", key,
+							),
+						}
+					}
+				},
+			},
```
afiune, line 66: Since we are adding a new resource that interacts with this Policy type, I don't think we should add this field into the resource schema.
```suggestion
```

afiune: Instead, just hardcode the type field on all API requests.

lacework/resource_lacework_policy_compliance.go
```diff
@@ -0,0 +1,264 @@
...
+	_, err := lacework.V2.Policy.Delete(d.Id())
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[INFO] Deleted Policy with guid %s\n", d.Id())
+	return nil
+}
+
+func importLaceworkPolicyCompliance(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
```
afiune, line 250: I think we should add a check to this import to verify that the Policy being imported is of type `Compliance`.

lacework/resource_lacework_policy_compliance.go
PengyuanZhao: Removed the type field and hardcoded it to Compliance.

lacework/resource_lacework_policy_compliance.go
PengyuanZhao: Added the check.

lacework/resource_lacework_policy_compliance.go
```diff
@@ -0,0 +1,255 @@
...
+	"errors"
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/lacework/go-sdk/api"
+)
+
+const policyType = "Compliance"
```
afiune, line 13: I love that you declared a constant here! Great job.

Though, we already expose one constant in the go-sdk, could we use it instead?

https://github.com/lacework/go-sdk/blob/main/api/policy.go#L56

You should be able to use it like `api.PolicyTypeCompliance.String()` below

lacework/resource_lacework_policy_compliance.go
```diff
@@ -0,0 +1,255 @@
...
+	log.Printf("[INFO] Importing Lacework Policy with guid: %s\n", d.Id())
+
+	response, err := lacework.V2.Policy.Get(d.Id())
+	if err != nil {
+		return nil, fmt.Errorf(
+			"Unable to import Lacework resource. Policy with guid '%s' was not found",
+			d.Id(),
+		)
+	}
+	if response.Data.PolicyType != "Compliance" {
```
afiune, line 247: For example, here we can do:
```suggestion
	if response.Data.PolicyType != api.PolicyTypeCompliance.String() {
```

afiune (changes requested):
a few more changes

lacework/resource_lacework_policy_compliance.go
```diff
@@ -0,0 +1,255 @@
...
+
+	policy := api.NewPolicy{
+		PolicyType:   policyType,
+		QueryID:      d.Get("query_id").(string),
+		Title:        d.Get("title").(string),
+		Enabled:      d.Get("enabled").(bool),
+		Description:  d.Get("description").(string),
+		Remediation:  d.Get("remediation").(string),
+		Severity:     d.Get("severity").(string),
+		AlertEnabled: d.Get("alerting.0.enabled").(bool),
```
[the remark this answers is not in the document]
PengyuanZhao: That makes sense! I changed it to `alerting_enabled`. But the violence policy still uses `alerting { enabled = ..., profile =... }`. Should we be concerned about the consistency?

afiune (commented):
This looooooks great! 

There is one more thing to do before merging this new resource, you need to add
documentation just like the `lacework_policy` resource:
* This doc https://github.com/lacework/terraform-provider-lacework/blob/main/website/docs/r/policy.html.markdown
* Translates to this https://registry.terraform.io/providers/lacework/lacework/latest/docs/resources/policy

Could you create a new one for the new `lacework_policy_compliance` resource?

You will have to add it to the index/left-menu-bar like https://github.com/lacework/terraform-provider-lacework/blob/main/website/lacework.erb#L322

website/docs/r/policy.html.markdown
```diff
@@ -11,7 +11,7 @@ description: |-
 Lacework provides a highly scalable platform for creating, customizing, and managing custom policies
 against any datasource that is exposed via the Lacework Query Language (LQL).
 
-For more information, see the [Policy Overview Documentation](https://docs.lacework.com/custom-policies-overview).
+For more information, see the [Policy Overview Documentation](https://docs.lacework.net/console/custom-policy-overview).
```
afiune, line 14: nice! Thank you

afiune (approved):
AMAZING!

PR #567, opened 2023-10-04
chore(CAD-508): Remove team-members from terraform provider

zekisherif:
***Issue***: https://lacework.atlassian.net/browse/LINK-1972

***Description:***
Since RBAC v2 has been enabled for all customers for quite a while. We don't need this API in documentation anymore as it will not work for any customer

This is also being down now because it was raised as an issue in https://lacework.atlassian.net/browse/LINK-1972

Related PR for API V2: https://github.com/lacework/rainbow/pull/14424

***Additional Info:***


zekisherif:
A resource group integration test is failing in the test-build pipeline due to a bug in rainbow that @lwmobeent has recently fixed. It's not related to this Pr and @afiune is aware of it.

lacework/provider.go
```diff
@@ -124,7 +124,6 @@ func Provider() *schema.Provider {
 			"lacework_resource_group_container":               resourceLaceworkResourceGroupContainer(),
 			"lacework_resource_group_gcp":                     resourceLaceworkResourceGroupGcp(),
 			"lacework_resource_group_machine":                 resourceLaceworkResourceGroupMachine(),
-			"lacework_team_member":                            resourceLaceworkTeamMember(),
```
nikhilesh-payyavuala, line 127: Do all your PR's need to go in any specific order of deployment? Or given no one is using it we should be fine?

nikhilesh-payyavuala (approved):

lacework/provider.go
zekisherif: @nikhilesh-payyavuala See conversation here https://lacework.slack.com/archives/C04CYEQ3W3G/p1696444342248359

Because TeamMembers breaks the moment we migrated all customers to v2 we are now in a bit of predicament.

@afiune outliend the standard deprecation process for go cli/terraform here https://github.com/lacework/go-sdk/pull/1398#pullrequestreview-1658326828 but we can't really do the last step as it's broken for every customer already.

PR #611, opened 2024-01-10
feat: add lacework metrics data source

dmurray-lacework:
* Requires api-server release

***Issue***: https://lacework.atlassian.net/browse/GROW-2627

***Description:***
Add new data source `lacework_metrics` to allow modules using the provider to send metrics to Lacework


lacework/data_source_lacework_metrics.go
```diff
@@ -0,0 +1,60 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The name of the module",
+			},
+			"version": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The version of the module",
+			},
+			"dataset": {
```
afiune, line 25: I would like users to modify this value, so let's not give them that control.

afiune (changes requested):

lacework/data_source_lacework_metrics.go
dmurray-lacework: Fixed!

lacework/provider.go
```diff
@@ -134,6 +134,7 @@ func Provider() *schema.Provider {
 			"lacework_api_token":          dataSourceLaceworkApiToken(),
 			"lacework_agent_access_token": dataSourceLaceworkAgentAccessToken(),
 			"lacework_user_profile":       dataSourceLaceworkUserProfile(),
+			"lacework_metrics":            dataSourceLaceworkMetrics(),
```
afiune, line 137: alphabetical order:
```suggestion
			"lacework_metrics":            dataSourceLaceworkMetrics(),
			"lacework_user_profile":       dataSourceLaceworkUserProfile(),
```

lacework/data_source_lacework_metrics.go
```diff
@@ -0,0 +1,50 @@
...
+		module_version = d.Get("version").(string)
+	)
+
+	honeycombEvent := api.NewHoneyvent(module_version, name, "lacework-terraform")
+	resp, err := lacework.V2.Metrics.Send(honeycombEvent)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(time.Now().UTC().String())
```
afiune, line 46: Could the trace_id be the datasource id? An honest question 🤔

examples/data_source_lacework_metrics/main.tf
```diff
@@ -0,0 +1,18 @@
...
+  required_providers {
+    lacework = {
+      source = "lacework/lacework"
+    }
+  }
+}
+
+provider "lacework" {}
+
+data "lacework_metrics" "test" {
```
afiune, line 11: nit: Should we make the data source name singular? The name in plural makes it sound we are collecting multiple metrics, but in reality is only one.

afiune (approved):
Since this is an "internal" data source, I am fine with the current implementation, my only concerns are:
* The name of the data source sounds very generic `lacework_metrics` but the metric itself is just for modules 🤔 
* The argument `name` is also generic, but the description says `"The name of the module"`

Are we planning to:
1. Add new arguments to this data source for other "features" or "metrics"?
2. Or, are we planning to add new data sources

PR #619, opened 2024-03-28
fix: adding resource tags to policy exception data type

ipcrm:
***Issue***: https://lacework.atlassian.net/browse/GROW-2831

***Description:***

Prior to this commit the data type supplied via the policy exception field_value_map resulted in map[string]string being sent to the API. The required format is map[string][]string. This commit fixes this.

***Additional Info:***
See example and test updates.


docs/resources/policy_exception.md
```diff
@@ -27,13 +27,36 @@ resource "lacework_policy_exception" "example" {
 }
 ```
 
+
+Create a Lacework Policy Exception to exempt specified resourceTags from policy.
```
afiune, line 31:
```suggestion
Create a Lacework Policy Exception to exempt specified `resourceTags` from policy.
```

docs/resources/policy_exception.md
```diff
@@ -27,13 +27,36 @@ resource "lacework_policy_exception" "example" {
...
+```
+
 ## Argument Reference
 
 The following arguments are supported:
 
 * `description` - (Required) The description of the policy exception.
 * `policy_id` - (Required) The id of the policy the exception is associated.
 * `constraint` - (Required) Constraint. See [Constraint](#Constraint) below for details.
+* `field_value_map` - (Optional) FieldValueMap. See[FieldValueMap](#FieldValueMap) below for details.
```
afiune, line 59: This doesn't go here. Move it to the next section (Constraint)

ipcrm: 👍 resolved in https://github.com/lacework/terraform-provider-lacework/pull/619/commits/d7d38e17661cb36798815e376acfadda914d6560

afiune (approved):

rwlodarczyk-xealth:
Looks great! Thanks for the fix!

ipcrm:
Unfortunately we cannot release this as-is because it will cause issues with existing state, and even with a state upgrader (tested) this will still cause errors because of the schema type change (string -> []string).

To avoid breakage, we are going to introduce a new `field_values_map` (plural instead of singular) that supports the correct data type. The existing `field_value_map` will be marked deprecated and removed next version bump.  @rwlodarczyk-xealth apologies on the delay on this, will get his wrapped hopefully in the next day.

ipcrm:
Revised the approach here; from the commit message:

```
Deprecates field_value_map for field_values_map. The data type needs to
change for the exceptions to take effect. By deprecating field_value_map
instead of changing it we will not introduce a breaking change, corrupt
any current state, and avoid any required code changes due to breakage.
```

afiune (approved):

PR #620, opened 2024-03-29
ci: add initial github workflows to replace codefresh

tmac1973:
***Issue***: 

https://lacework.atlassian.net/browse/GROW-2760

***Description:***

These are initial versions of the workflows to replace codefresh. These are likely not final but I need some version of these in main to be able to execute them in my branch. These are currently set to manual execution only.

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -1,3 +1,5 @@
+name: Prepare Release
+
 on: workflow_dispatch
 
 jobs:
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -1,0 +1,53 @@
+on: workflow_dispatch
+
+jobs:
+  prepare-release:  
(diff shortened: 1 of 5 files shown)


.github/workflows/test-build.yml
```diff
@@ -0,0 +1,155 @@
+name: Nightly Build
```
PengyuanZhao, line 1:
```suggestion
name: Test Build
```

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,53 @@
+on: workflow_dispatch
```
PengyuanZhao, line 1: nit: define a name. Same for the other workflows.

PengyuanZhao (approved):

.github/workflows/prepare-release.yml
tmac1973: Will do in next PR

PR #624, opened 2024-04-05
ci: migrate from codefresh to github workflows

tmac1973:
***Issue***: 
https://lacework.atlassian.net/browse/GROW-2760

***Description:***
Provides workflows for testing and release to replace codefresh

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -77,8 +77,9 @@ jobs:
           LW_API_KEY: ${{ secrets.LW_API_KEY }}
           LW_API_SECRET: ${{ secrets.LW_API_SECRET }}
           LW_SUBACCOUNT: ${{ secrets.LW_SUBACCOUNT }}
-          OCI_PRIVATE_KEY: ${{ secrets.OCI_CREDENTIALS }}
+          OCI_CREDENTIALS: ${{ secrets.OCI_CREDENTIALS }}
           GOOGLE_CREDENTIALS: ${{ secrets.GOOGLE_CREDENTIALS }}
+          AZURE_CREDENTIALS: ${{ secrets.AZURE_CREDENTIALS }}
 
   slack-notify:
     name: Slack Notify if Failed Tests
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -77,6 +77,8 @@ jobs:
           LW_API_KEY: ${{ secrets.LW_API_KEY }}
           LW_API_SECRET: ${{ secrets.LW_API_SECRET }}
           LW_SUBACCOUNT: ${{ secrets.LW_SUBACCOUNT }}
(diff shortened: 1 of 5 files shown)


.github/workflows/nightly.yml
```diff
@@ -2,8 +2,8 @@ name: Nightly Build
 
 on:
   workflow_dispatch:
```
ipcrm, line 4: all of these have workflow_dispatch; 

This is required for recovery purposes, correct?

.github/workflows/nightly.yml
```diff
@@ -72,14 +72,31 @@ jobs:
           make prepare
           make clean-test
           make integration-test
+        env:
+          LW_ACCOUNT: ${{ secrets.LW_ACCOUNT }}.lacework.net
```
ipcrm, line 76: secret is shortname? (I assume)

ipcrm (commented):
Looks good! couple q's

.github/workflows/nightly.yml
tmac1973: This allows us to run the jobs manually if we like, which provides parity with codefresh

.github/workflows/test-build.yml
```diff
@@ -1,9 +1,11 @@
 name: Test Build
 
-on:
+on: 
   workflow_dispatch:
-  # pull_request:
-  # push:
+  pull_request:
```
ipcrm, line 5: do we want to run for ANY pull request or only prs targeting main?

.github/workflows/test-build.yml
```diff
@@ -72,14 +74,31 @@ jobs:
...
+          AWS_S3: ${{ secrets.AWS_S3 }}
+          S3_BUCKET_ARN: ${{ secrets.S3_BUCKET_ARN }}
+          LW_PUB_SUB_SUBSCRIPTION: ${{ secrets.LW_PUB_SUB_SUBSCRIPTION }}
 
   slack-notify:
     name: Slack Notify if Failed Tests
     needs: [run-unit-tests, build, run-integration-tests]
     runs-on: ubuntu-latest
-    if: needs.run-unit-tests.result == 'failure' || needs.build.result == 'failure' || needs.run-integration-tests.result == 'failure'
+    if: false # always()
```
ipcrm, line 98: I think this should be uncommented 

```suggestion
    if: always()
```

ipcrm (commented):
Just a couple questions!

.github/workflows/test-build.yml
tmac1973: I don't see a problem running the test for any PR. Do you have a reason to restrict it to main?

ipcrm: Not really tbh; I just know the historic behavior was only build PRs targeting main. But that said, if you don't see an issue with it; neither do I!

tmac1973: changed to main only for PR

ipcrm (approved):

PR #625, opened 2024-04-08
chore: add the optional `queryLanguage` enum value where used

gspofford-lw:
***Issue***: https://lacework.atlassian.net/browse/RAIN-93039

***Description:***
This change adds the `queryLanguage` enumeration to the lacework_query and lacework_policy_compliance resource types.

For a query, it is an optional part of the create submission. It is also an optional part of query payload response from the result of creating, updating, or retrieving the definition of a query.

For a policy, it is an optional part of the response from creating, updating or retrieving the definition of a policy.

***Additional Info:***

### DIFF
--- a/lacework/resource_lacework_policy_compliance.go
+++ b/lacework/resource_lacework_policy_compliance.go
@@ -135,7 +135,9 @@ func resourceLaceworkPolicyComplianceCreate(d *schema.ResourceData, meta interfa
 	}
 
 	d.SetId(response.Data.PolicyID)
-	d.Set("query_language", response.Data.QueryLanguage)
+	if response.Data.QueryLanguage != nil {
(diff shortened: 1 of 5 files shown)


lacework/resource_lacework_policy_compliance.go
```diff
@@ -33,6 +33,13 @@ func resourceLaceworkPolicyCompliance() *schema.Resource {
 				Required:    true,
 				Description: "The id of the query",
 			},
+			"query_language": {
+				Type:        schema.TypeString,
+				Required:    false,
+				Optional:    true,
+				Computed:    true,
```
PengyuanZhao, line 40: This computed value needs to be set with `d.Set("query_language", response.Data.QueryLanguage)` too I think.

PR #654, opened 2024-09-10
fix(LINK-3033): set query_id when importing lacework_resource_query

PengyuanZhao:
***Issue***: https://lacework.atlassian.net/browse/LINK-3033

***Description:***
The issue is when importing lacework_resource_query, the `query_id` is not set which causes a state drift. This PR fixes it.


leijin-lw (approved):
PR looks good. However, there are failing tests with resource group tests. We can disable the rgv1 tests

PR #725, opened 2026-05-15
chore: run gofumpt formatter against codebase

RoseSecurity:
***Description:***
This PR runs `gofumpt` across the Go codebase to normalize formatting and bring existing files in line with stricter Go formatting conventions.

The changes are formatting-only and include updates such as multi-line function call formatting, trailing commas in multi-line argument lists, simplified single-variable declarations, and removal of unnecessary blank lines. No provider behavior or resource logic is intended to change.

***Additional Info:***
This touches 70 Go files across `lacework/`, `integration/`, and `main.go`.

Tests were not run locally because this is a formatting-only change.


PengyuanZhao:
Thanks for this, @RoseSecurity, and sorry it sat open so long.

Closing for now — not because the change is wrong, but because nothing would keep it true. gofmtcheck.sh runs plain gofmt, .golangci.yaml configures no linters, and make lint isn't wired into CI, so gofumpt style would drift with the next commit and we'd have taken 70 files of churn for it.

We will look into this formatting issue later.

PR #726, opened 2026-06-09
feat(AWLS2-1027): org/tenant level support for DSPM integrations

wilderj:
## Jira Ticket
https://lacework.atlassian.net/browse/AWLS2-1027

## Summary

These changes are needed for tenant/organization level support in DSPM. See detailed description in the matching go-sdk PR: https://github.com/lacework/go-sdk/pull/1835

This PR adds code to our Terraform provider to actually create these tenant/org level integrations using the new fields introduced elsewhere. 

Right now this is blocked on the go-sdk changes, and won't compile until those changes are merged and the dependencies bumped.


lacework/resource_lacework_aws_dspm.go
```diff
@@ -68,6 +68,16 @@ func resourceLaceworkAwsDspm() *schema.Resource {
 					Type: schema.TypeString,
 				},
 			},
+			"integration_level": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Deployment scope: 'ACCOUNT' (single account) or 'ORG' (whole organization). Defaults to ACCOUNT.",
+			},
```
kirklandnuts, line 94: we should have the same hardened schema (like validation and whatnot) here as you have for azure integration_level:
```
			"integration_level": {
				Type:     schema.TypeString,
				Optional: true,
				Default:  api.AzureSubscriptionIntegration,
				StateFunc: func(val interface{}) string {
					return strings.ToUpper(val.(string))
				},
				ValidateFunc: func(value interface{}, key string) ([]string, []error) {
					switch strings.ToUpper(value.(string)) {
					case api.AzureSubscriptionIntegration,
						api.AzureTenantIntegration:
						return nil, nil
					default:
						return nil, []error{
							fmt.Errorf("%s: can only be either '%s' or '%s'",
								key,
								api.AzureSubscriptionIntegration,
								api.AzureTenantIntegration),
						}
					}
				},
				Description: "Integration level - TENANT / SUBSCRIPTION.",
			},
```

lacework/resource_lacework_aws_dspm.go
```diff
@@ -128,15 +161,17 @@ func resourceLaceworkAwsDspmCreate(d *schema.ResourceData, meta interface{}) err
 			ExternalID: d.Get("credentials.0.external_id").(string),
 			RoleArn:    d.Get("credentials.0.role_arn").(string),
 		},
-		Regions: castAttributeToStringSlice(d, "regions"),
+		Regions:           castAttributeToStringSlice(d, "regions"),
+		IntegrationLevel:  d.Get("integration_level").(string),
```
kirklandnuts: I don't think it's as important if we define the rigid schema as I mentioned in my other comment, but we should treat integration level here the same as we do for azure (with explicit default and uppercasing): 
```
	integrationLevel := api.AzureSubscriptionIntegration
	if strings.ToUpper(d.Get("integration_level").(string)) == api.AzureTenantIntegration {
		integrationLevel = api.AzureTenantIntegration
	}
```

kirklandnuts (approved):
I think we should treat integration values for aws and azure consistently - otherwise, this LGTM

lacework/resource_lacework_dspm_helpers.go
```diff
@@ -52,6 +57,25 @@ func buildDspmProps(d *schema.ResourceData) (*api.DspmProps, error) {
...
 
+	if v, ok := d.GetOk(accountFilterKey); ok {
+		filters := v.([]interface{})
+		if len(filters) > 0 && filters[0] != nil {
+			filterMap := filters[0].(map[string]interface{})
+			filterMode := filterMap["filter_mode"].(string)
+			ids := castAndTransformStringSlice(filterMap[accountIdsKey].([]interface{}), func(s string) string { return s })
+			if len(ids) == 0 {
+				return nil, fmt.Errorf("%s is required when %s.filter_mode is '%s'", accountIdsKey, accountFilterKey, filterMode)
+			}
```
kirklandnuts: Isn't it valid for ids to be empty list when filterMode is "ALL"?
```suggestion
			if filterMode != "ALL" && len(ids) == 0 {
				return nil, fmt.Errorf("%s is required when %s.filter_mode is '%s'", accountIdsKey, accountFilterKey, filterMode)
			}
```

kirklandnuts (changes requested):
Actually, I think there's an issue: the code currently doesn't have a path for `ALL` filter mode in which case we'd expect accountIds to be an empty list.

lacework/resource_lacework_aws_dspm.go
wilderj: good call, done

lacework/resource_lacework_aws_dspm.go
wilderj: yep totally fair.

lacework/resource_lacework_dspm_helpers.go
wilderj: Great catch ... I have not tested that path yet. Fixed.

kirklandnuts (approved):

lacework/resource_lacework_aws_dspm.go
```diff
@@ -111,6 +140,29 @@ func resourceLaceworkAwsDspm() *schema.Resource {
...
+				Type:        schema.TypeList,
+				Optional:    true,
+				MaxItems:    1,
+				Description: "Narrow an org-level scan to specific accounts (only meaningful when integration_level is 'ORG').",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"filter_mode": {
+							Type:        schema.TypeString,
+							Required:    true,
+							Description: "Filter mode: 'INCLUDE' or 'EXCLUDE'.",
```
kirklandnuts, line 153:
```suggestion
							Description: "Filter mode: 'INCLUDE', 'EXCLUDE', or 'ALL'.",
```

lacework/resource_lacework_azure_dspm.go
```diff
@@ -35,6 +36,52 @@ func resourceLaceworkAzureDspm() *schema.Resource {
...
+				Type:        schema.TypeList,
+				Optional:    true,
+				MaxItems:    1,
+				Description: "For TENANT integrations, narrow which subscriptions are scanned. Stored on the integration props, mirroring datastore_filters.",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"filter_mode": {
+							Type:        schema.TypeString,
+							Required:    true,
+							Description: "Filter mode: 'INCLUDE' or 'EXCLUDE'.",
```
kirklandnuts, line 72:
```suggestion
							Description: "Filter mode: 'INCLUDE', 'EXCLUDE', or 'ALL'.",
```

lacework/resource_lacework_azure_dspm.go
```diff
@@ -35,6 +36,52 @@ func resourceLaceworkAzureDspm() *schema.Resource {
...
+						},
```
kirklandnuts, line 73: nit: we should validate `filter_mode` using a `ValidateFunc` like we do for `integration_level`

PR #752, opened 2026-08-27
fix for CAD-1812, validate lql query in terraform

lathaftnt:
***Issue***: 
https://lacework.atlassian.net/browse/CAD-1812

***Description:***
AWS Agentless Integration will not complain about mis-configured LQL in terraform. Terraform applied a misconfigured LQL statement successfully without any error messages or warnings.  This fix adds terraform validation.

***Additional Info:***
I built and installed the terraform provider locally and executed the valid and invalid lql queries against it to cover the following cases:
1. Error returned with an invalid lql query (with all types of providers)
2. Success returned with a valid lql query (with all types of providers)


kirklandnuts (approved):
this looks good, thanks Latha

lacework/resource_lacework_integration_agentless_scanning_query.go
lokesh-vadlamudi, line 1: may be change the file name to agentless_scanning_helper.go? this current filename sounds like a resource instead of a helper.  (eg. account_mapping_helper.go)

lacework/resource_lacework_integration_agentless_scanning_query.go
```diff
@@ -0,0 +1,29 @@
...
+//
+// The integration API itself does not validate query_text, so without this
+// check a syntactically invalid query (e.g. unbalanced braces, or a
+// nonexistent LQL function like is_null()) is silently persisted, and the
+// scanner falls back to scanning everything with no error surfaced anywhere.
+//
+// An empty query_text means "scan everything" and is intentionally not
+// validated.
+func validateAgentlessScanningQueryText(lacework *api.Client, queryText string) error {
+	if queryText == "" {
```
lokesh-vadlamudi: add wildcard support (since that also means scan all). || queryText == "*"

lokesh-vadlamudi, line 21: add wildcard support (since that also means scan all). || queryText == "*"

lacework/resource_lacework_integration_agentless_scanning_query.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+// validateAgentlessScanningQueryText validates the LQL filter query used by
+// Limited Workload Scanning before it is sent to the Lacework cloud account
+// integration API.
+//
+// The integration API itself does not validate query_text, so without this
+// check a syntactically invalid query (e.g. unbalanced braces, or a
+// nonexistent LQL function like is_null()) is silently persisted, and the
+// scanner falls back to scanning everything with no error surfaced anywhere.
```
marktabry, line 16: The first clause is right, the second isn't. On the api-server only empty-string and `*` short-circuit to scan-everything (`AgentlessSidekickHelper.java:190` -> `{"ALL":1}`); a non-empty invalid query falls through to `lqlQueries.executeQuery(...)` at `:230-231` with no catch and no fallback. So the failure mode is a failing query execution, not a silent scan-all.

The rationale for the PR is unaffected -- an invalid query being silently *persisted* is the real problem, and that part is accurate. Worth trimming the "falls back to scanning everything" clause so the next reader isn't hunting for a fallback that doesn't exist.

lacework/resource_lacework_integration_agentless_scanning_query.go
```diff
@@ -0,0 +1,29 @@
...
+//
+// An empty query_text means "scan everything" and is intentionally not
+// validated.
+func validateAgentlessScanningQueryText(lacework *api.Client, queryText string) error {
+	if queryText == "" {
+		return nil
+	}
+
+	if _, err := lacework.V2.Query.Validate(api.ValidateQuery{QueryText: queryText}); err != nil {
+		return fmt.Errorf("invalid query_text: %s", err)
```
marktabry, line 26: Every non-2xx gets relabelled as `invalid query_text` -- a 403, a timeout, a DNS failure all come back telling the practitioner their query is bad. go-sdk's own CLI words this as `"unable to validate query"` (`cli/cmd/lql_validate.go:30`) for exactly this reason; suggest `failed to validate query_text: %s`.

Related: this call isn't retried. `api.WithRetries` is never passed to the client (`lacework/provider.go:153-200`), so it runs exactly once -- while all four Create paths wrap the cloud-account call immediately after it in `retry.RetryContext` with a user-tunable `retries`. A blip the create would have ridden out now aborts the apply and blames the query.

lacework/resource_lacework_integration_aws_agentless_scanning.go
```diff
@@ -265,6 +269,10 @@ func resourceLaceworkIntegrationAwsAgentlessScanningUpdate(d *schema.ResourceDat
 		awsAgentlessScanningData.QueryText = d.Get("query_text").(string)
 	}
 
+	if err := validateAgentlessScanningQueryText(lacework, awsAgentlessScanningData.QueryText); err != nil {
```
marktabry, line 272: Unconditional on Update, so a config that still declares a `query_text` the API previously accepted now fails `terraform apply` on any unrelated change -- bumping `scan_frequency`, toggling `enabled`. Those configs are the population this PR exists for, so it's worth deciding deliberately rather than by omission. Same pattern at `_aws_org_:375`, `_azure_:342`, `_gcp_:364`.

`if d.HasChange("query_text") { ... }` on the Update side keeps the intent (you can't introduce or edit a bad query) while letting an existing one be fixed on purpose. Create stays unconditional. Minor either way -- the remedy is a one-line config edit.

lacework/resource_lacework_integration_agentless_scanning_query.go
```diff
@@ -0,0 +1,29 @@
...
+// scanner falls back to scanning everything with no error surfaced anywhere.
+//
+// An empty query_text means "scan everything" and is intentionally not
+// validated.
+func validateAgentlessScanningQueryText(lacework *api.Client, queryText string) error {
+	if queryText == "" {
+		return nil
+	}
+
+	if _, err := lacework.V2.Query.Validate(api.ValidateQuery{QueryText: queryText}); err != nil {
```
marktabry, line 25: Heads-up on a permission change: `POST api/v2/Queries/validate` resolves to entity `Policies`, action `read` (`uriToEntityMapping.json:100` matches the `Queries/(...(?<!execute)$)` pattern; `RbacHelper.java:220` maps POST-ending-in-`/validate` to `read`). Managing these integrations previously needed only `CloudAccounts:write` (`uriToEntityMapping.json:43-44`).

Narrow in practice -- all built-in roles that can write cloud accounts already carry `Policies:read`, org admins bypass, and it's behind `RBAC_V2_FEATURE_FLAG`. But it's reachable for a customer-defined least-privilege role like `{"CloudAccounts":["read","write"]}`, which is the shape a terraform service account tends to get, and it surfaces only as a mid-apply 403. A line in the resource docs / CHANGELOG probably covers it.

marktabry (commented):
Reviewed the validation helper and its eight call sites. Nothing here blocks merge -- the change does what it sets out to do, and it brings terraform in line with the console, which has run the same validation since CAD-1734. Five minor notes, of which the first two are worth fixing before merge. I also dug into the wildcard question on Lokesh's thread and replied there with what I found.

lacework/resource_lacework_integration_agentless_scanning_query.go
marktabry: Chased this one down, and I think the exemption would be the wrong call -- flagging so it's a deliberate decision rather than a reflex.

You're right that the backend maps `*` to scan-all: `AgentlessSidekickHelper.java:190` short-circuits `query.length() == 0 || query.equals(SIDEKICK_QUERY_WILDCARD)` to `{"ALL":1}`. But `*` doesn't appear to be a supported user-facing value:

- `SIDEKICK_QUERY_WILDCARD` has exactly two occurrences in the whole `services` monorepo -- its definition at `:58` and that one use. Undocumented 2022 defensive branch.
- Every documented surface uses `""`: all four `examples/*/main.tf`, `terraform-aws-agentless-scanning`'s `filter_query_text` ("If left blank, Lacework will scan all resources"), and the console help text.
- The console already rejects `*`. `CloudAccountIntegrationEdit.tsx:118` gates on `if (queryText.trim())`, and `'*'.trim()` is truthy -- so `*` hits the same validate endpoint and fails. Shipped since Dec 2025 under CAD-1734.
- And validate does reject it: the grammar's `query` rule requires a leading `{` (`LQL2.g4:41-44`), `LQL2ErrorListener` throws on first syntax error, mapped to 400.

So adding `|| queryText == "*"` would make terraform more permissive than the console. Happy to be overruled if `*` is contractual somewhere I couldn't find.

lacework/agentless_scanning_helper.go
lathaftnt: I have made this change

lacework/agentless_scanning_helper.go
```diff
@@ -0,0 +1,29 @@
...
+//
+// The integration API itself does not validate query_text, so without this
+// check a syntactically invalid query (e.g. unbalanced braces, or a
+// nonexistent LQL function like is_null()) is silently persisted, and the
+// scanner falls back to scanning everything with no error surfaced anywhere.
+//
+// An empty query_text means "scan everything" and is intentionally not
+// validated.
+func validateAgentlessScanningQueryText(lacework *api.Client, queryText string) error {
+	if queryText == "" {
```
lathaftnt: I have not made this change based on the above comments.

lacework/agentless_scanning_helper.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+// validateAgentlessScanningQueryText validates the LQL filter query used by
+// Limited Workload Scanning before it is sent to the Lacework cloud account
+// integration API.
+//
+// The integration API itself does not validate query_text, so without this
+// check a syntactically invalid query (e.g. unbalanced braces, or a
+// nonexistent LQL function like is_null()) is silently persisted, and the
+// scanner falls back to scanning everything with no error surfaced anywhere.
```
lathaftnt: I have made this change

lacework/agentless_scanning_helper.go
```diff
@@ -0,0 +1,29 @@
...
+//
+// An empty query_text means "scan everything" and is intentionally not
+// validated.
+func validateAgentlessScanningQueryText(lacework *api.Client, queryText string) error {
+	if queryText == "" {
+		return nil
+	}
+
+	if _, err := lacework.V2.Query.Validate(api.ValidateQuery{QueryText: queryText}); err != nil {
+		return fmt.Errorf("invalid query_text: %s", err)
```
lathaftnt: I have not added the retries, this has to be done based on return code, else a genuine invalid query will retry and introduce latency, If this fails due to a transient network issue rather than the query itself, simply re-run apply would be a better option.

lacework/resource_lacework_integration_aws_agentless_scanning.go
lathaftnt: I have made this change

lacework/agentless_scanning_helper.go
```diff
@@ -0,0 +1,29 @@
...
+// scanner falls back to scanning everything with no error surfaced anywhere.
+//
+// An empty query_text means "scan everything" and is intentionally not
+// validated.
+func validateAgentlessScanningQueryText(lacework *api.Client, queryText string) error {
+	if queryText == "" {
+		return nil
+	}
+
+	if _, err := lacework.V2.Query.Validate(api.ValidateQuery{QueryText: queryText}); err != nil {
```
lathaftnt: I have added note in the resource docs. I didn't find one for Azure though, will be good if someone can help create it please.

lokesh-vadlamudi (approved):
LGTM

wsli-1 (approved):

marktabry (approved):

PR #757, opened 2026-09-14
feat: add lacework_integration_{aws,azure}_fortidspm resources

cstrace-ftnt:
***Issue***: https://lacework.atlassian.net/browse/DSPM-13

***Description:***

**What**

Three resources for DSPM cloud accounts that FortiDSPM manages itself, plus the go-sdk bump (v2.15.1 → v2.17.0, re-vendored) that carries their API types.

- `lacework_integration_aws_fortidspm` / `lacework_integration_azure_fortidspm` create a DSPM cloud account (`AwsDspm` / `AzureDspm` type) with only the account identity and the regions: AWS account id, or Azure tenant id and optional subscription id. There is no customer bucket and no cross-account credentials — FortiDSPM pushes scan results to Lacework itself. On create, api-server asks FortiDSPM for a deployment and the resource exposes it as computed attributes keyed by region, which is the shape Terraform `for_each` needs: `deployment_id`, `activation_tokens` (sensitive), `image_ids` (AWS AMI ids), `image_urls` (sensitive, Azure signed VHD URLs), `hyperv_generations`, `token_expires_in`, `image_url_expires_in`.
- `lacework_fortidspm_deployment_status` reports the Terraform outcome for one integration to FortiDSPM through `PUT /api/v2/FortiDspm/deployments/{intgGuid}/status`: overall status, per-region status with the scan engine's instance id / IPs / role, and an optional error. Create and Update both report; Read and Delete are no-ops because the report is write-only and the integration's own delete tells FortiDSPM the deployment is gone.

**How**

The two integration resources follow `lacework_integration_aws_dspm` / `lacework_integration_azure_dspm`: same CRUD shape, same `retries` / `enabled` handling, `CloudAccounts.CreateAwsFortiDspm` / `CreateAzureFortiDspm` / `GetAwsFortiDspm` / `GetAzureFortiDspm` from go-sdk. The deployment is issued once at create time and is not returned by later reads, so Read only refreshes the identity fields and leaves the deployment attributes in state. The status resource's arguments reference the scan engine resources, so Terraform orders it after they exist; an apply that fails earlier never reports, which is the agreed behaviour with FortiDSPM.

**Notable**

- The existing `lacework_integration_aws_dspm` / `lacework_integration_azure_dspm` resources are unchanged.
- go-sdk v2.17.0 (lacework/go-sdk#1861) is the only dependency change; `go mod vendor` added `api/cloud_accounts_fortidspm.go` and touched `api/api.go`, `api/version.go` and `modules.txt`.
- Docs for the three resources are under `docs/resources/`.

***Additional Info:***

- `go build ./...`, `make fmtcheck` and the unit tests pass against the released v2.17.0 (no local replace).
- End to end with a provider built from this branch: a customer-style `terraform apply` of `terraform-azure-fortidspm` created the integration through `lacework_integration_azure_fortidspm`, copied the VHD from the returned `image_urls`, built the image and VM, and `lacework_fortidspm_deployment_status` reported `succeeded` for `westus2`; `terraform destroy` removed the resources and the integration. The same provider drove the Cloud Accounts wizard through self-deployment for AWS (`us-west-2`, scan engine activated against FortiDSPM) and Azure.
- The `terraform-{aws,azure}-fortidspm` modules that consume these resources are in separate PRs and pin `lacework >= 2.6.0` once this ships.


PengyuanZhao (approved):

lacework/resource_lacework_fortidspm_deployment_status.go
```diff
@@ -0,0 +1,147 @@
...
+			Region:              r["name"].(string),
+			Status:              r["status"].(string),
+			InstanceID:          r["instance_id"].(string),
+			PrivateIP:           r["private_ip"].(string),
+			NatGatewayPublicIP:  r["nat_gateway_public_ip"].(string),
+			IamRoleArn:          r["iam_role_arn"].(string),
+			IdentityPrincipalID: r["identity_principal_id"].(string),
+		})
+	}
+	if msg := d.Get("error_message").(string); msg != "" {
```
lokesh-vadlamudi: nit:
is this a better alternative?
phase, msg := d.Get("error_phase").(string), d.Get("error_message").(string)
if phase != "" || msg != "" {
        report.Error = &api.FortiDspmDeploymentError{Phase: phase, Message: msg}
}

lokesh-vadlamudi (commented):
LGTM

lacework/resource_lacework_fortidspm_deployment_status.go
cstrace-ftnt: Good call, adopted as suggested in 9693d09.

lokesh-vadlamudi (approved):

